PR #11, opened 2021-10-21
Update to NPM v7

tpmcgowan:


package.json
```diff
@@ -136,7 +136,7 @@
     "concurrently": "^6.3.0",
     "cypress": "^8.6.0",
     "cypress-multi-reporters": "^1.5.0",
-    "eslint": "^7.3.2",
+    "eslint": "^7.32.0",
```
psoleckimoj, line 139: Confused at this? Latest is 8.0.1? But can't use that as other dependencies require 7.3.2.

PR #13, opened 2021-11-15
VB-230: Add initial tiles

psoleckimoj:
Contains some basic tests, although this is a very basic page layout.

### DIFF
--- a/assets/js/card.js
+++ b/assets/js/card.js
@@ -1,0 +1,12 @@
+window.onload = function () {
+  // Loops through dom and finds all elements with card--clickable class
+  document.querySelectorAll('.card--clickable').forEach(card => {
+    // Check if card has a link within it
+    if (card.querySelector('a') !== null) {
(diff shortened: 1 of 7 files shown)


tpmcgowan:
Looks good! Just looking at the header, the one in the MoJ pattern library (https://design-patterns.service.justice.gov.uk/components/header/) has blue bar at the bottom unlike the one pulled in from the template app. Should we be using that? (I tried swapping out contents of header.njk and it worked OK...)

psoleckimoj:
> Looks good! Just looking at the header, the one in the MoJ pattern library (https://design-patterns.service.justice.gov.uk/components/header/) has blue bar at the bottom unlike the one pulled in from the template app. Should we be using that? (I tried swapping out contents of header.njk and it worked OK...)

@tpmcgowan I don't think so, the DPS/template one is an update from GDS.

tpmcgowan (approved):
👍

PR #31, opened 2021-12-23
VB-314 Prisoner profile page

tpmcgowan:


server/data/prisonApiClient.ts
```diff
@@ -0,0 +1,38 @@
...
+  getOffender(offenderNo: string): Promise<InmateDetail> {
+    return this.restclient.get({
+      path: `/api/offenders/${offenderNo}`,
+    })
+  }
+
+  async getVisitBalances(offenderNo: string): Promise<VisitBalances | null> {
+    let balances: VisitBalances
+    try {
+      balances = await this.restclient.get({
```
psoleckimoj, line 24: Could always `return` here, then change the last `return` on line 34 to just `return null`

psoleckimoj: Then no need for `balances`

tpmcgowan: Good suggestion, thanks 👍!

assets/sass/components/_prisoner-profile.scss
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  &--RCDR {
+    background-color: govuk-colour("dark-blue");
+  }
+  
+  &--URCU {
+    background-color: govuk-colour("purple");
+  }
+}
```
psoleckimoj, line 52: needs a newline 👍

psoleckimoj (dismissed):
LGTM! Just wondering if we just have HTML in that type? It doesn't stop us just having text and may make it less complicated? We're also only using HTML now.

assets/sass/components/_prisoner-profile.scss
tpmcgowan: Keep forgetting those 🤦‍♂️!

tpmcgowan:
👍 will try out the type suggestion...

tpmcgowan:
Have added a change to simplify the type for the search result table row items to just have `html` rather than `html` or `text`. Looking in the govuk table macro it seems only difference is whether they're escaped or not so I think it'll be fine given how we're currently using it.

psoleckimoj (approved):
LGTM :)

PR #38, opened 2022-01-10
VB-314 Make extra API call to get convicted status on profile page

tpmcgowan:
Also restricts viewing of offenders to those in Hewell

### DIFF
--- a/server/services/prisonerProfileService.test.ts
+++ b/server/services/prisonerProfileService.test.ts
@@ -1,6 +1,7 @@
+import { NotFound } from 'http-errors'
 import PrisonerProfileService from './prisonerProfileService'
 import PrisonApiClient from '../data/prisonApiClient'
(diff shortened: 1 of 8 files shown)


server/services/prisonerProfileService.test.ts
```diff
@@ -25,6 +26,23 @@ describe('Prisoner profile service', () => {
     })
 
     it('Retieves and processes data for prisoner profile with visit balances', async () => {
```
psoleckimoj, line 28: I know it's not in the change but spelling 😁

psoleckimoj (dismissed):
LGTM

server/services/prisonerProfileService.test.ts
tpmcgowan: well spotted 🤦‍♂️!

psoleckimoj (approved):
lgtm

PR #44, opened 2022-01-13
VB-321: Sort VO dates

psoleckimoj:


server/services/prisonerProfileService.ts
```diff
@@ -27,13 +28,21 @@ export default class PrisonerProfileService {
     let visitBalances = null
 
     if (convictedStatus !== 'Remand') {
-      visitBalances = await prisonApiClient.getVisitBalances(offenderNo)
+      visitBalances = (await prisonApiClient.getVisitBalances(offenderNo)) as BAPVVisitBalances
 
       if (visitBalances.latestIepAdjustDate) {
+        visitBalances.nextIepAdjustDate = format(
```
psoleckimoj, line 34: I want to test this but at the same time, it's just testing a library which is already tested 🤷‍♂️

server/utils/utils.ts
```diff
@@ -5,8 +5,10 @@ const properCase = (word: string): string =>
 
 const isBlank = (str: string): boolean => !str || /^\s*$/.test(str)
 
+export const getDateFromAPI = (dateString: string): Date => parse(dateString, 'yyyy-MM-dd', new Date())
```
psoleckimoj, line 8: Created as we're likely to need to convert a date from endpoints quite frequently.

tpmcgowan (approved):
LGTM!

PR #46, opened 2022-01-19
Make function accept wrap param, put no wrapping in relevant places, …

psoleckimoj:
…update tests

NOTE: I haven't put a class on the table cells because we only want part of the date without wrapping.


tpmcgowan (approved):
LGTM!

PR #51, opened 2022-01-25
Replace express-request-id

psoleckimoj:


.eslintrc.json
```diff
@@ -93,7 +93,7 @@
         "functions": "never"
       }
     ],
-    "import/no-extraneous-dependencies": ["error", { "devDependencies": ["**/*.test.js", "**/*.test.ts"] }],
+    "import/no-extraneous-dependencies": ["error", { "devDependencies": true }],
```
psoleckimoj, line 96: Couldn't work out why it only complained about the type for uuid and no other modules?

server/@types/bapv.d.ts
```diff
@@ -1,5 +1,13 @@
 import { InmateDetail, VisitBalances } from '../data/prisonApiTypes'
 
+declare global {
```
psoleckimoj, line 3: Needed to allow adding `id` to the request object.

psoleckimoj:
@andrewrlee this is a quick draft of trying to pull the code into our own repo. Not tested, just thought I'd get your thoughts 🙂

server/@types/bapv.d.ts
andrewrlee: would it be better to add the type override [here](https://github.com/ministryofjustice/book-a-prison-visit-staff-ui/blob/9b21c7f7076e16184404856b11404881af2cc90a/server/%40types/express/index.d.ts#L19)

.eslintrc.json
andrewrlee: not sure this is right? Doesn't the change says that it shouldn't complain if you use a dev import in any code including prod? 

I think you just need to install `uuid` explicitly as a prod dependency?

server/@types/bapv.d.ts
psoleckimoj: Done 👍

psoleckimoj:
@andrewrlee I've made all the changes and all seems to be working well on dev. @tpmcgowan this sorts one of the issues with updating the `express-request-id` module by removing it 😁

tpmcgowan (approved):
LGTM

PR #52, opened 2022-01-25
Enable prisoner-contact-registry API endpoint

tpmcgowan:
Enable the prisoner-contact-registry endpoint rather than using hard-coded data. Currently this is only available on dev, so will break things on preprod/prod (but we've separately identified that we could better handle errors - this could be the chance!)


helm_deploy/values-preprod.yaml
```diff
@@ -16,7 +16,7 @@ generic-service:
     PRISONER_SEARCH_API_URL: "https://prisoner-offender-search-preprod.prison.service.justice.gov.uk"
     PRISON_API_URL: "https://api-preprod.prison.service.justice.gov.uk"
     VISIT_SCHEDULER_API_URL: "https://visit-scheduler-preprod.hmpps.service.justice.gov.uk"
-    PRISONER_CONTACT_REGISTRY_API_URL: "https://visit-scheduler-dev.hmpps.service.justice.gov.uk"
+    PRISONER_CONTACT_REGISTRY_API_URL: "https://prisoner-contact-registry-preprod.hmpps.service.justice.gov.uk"
```
tpmcgowan, line 19: this (and the prod one) aren't working yet but putting what I'm assuming they'll be so they're ready

server/data/prisonerContactRegistryApiClient.ts
```diff
@@ -15,16 +14,12 @@ class PrisonerContactRegistryApiClient {
   constructor(private readonly restclient: RestClient) {}
 
   getPrisonerSocialContacts(offenderNo: string): Promise<Contact[]> {
-    if (config.apis.prisonerContactRegistry.enabled) {
```
tpmcgowan, line 18: figured I may as well remove the toggle and test data - though could leave it in place if you prefer?

helm_deploy/values-preprod.yaml
psoleckimoj: 👍

server/data/prisonerContactRegistryApiClient.ts
psoleckimoj: Happy to remove 👍

psoleckimoj (approved):
LGTM

PR #56, opened 2022-02-01
Select visitors from the prisoner's visitor list

tpmcgowan:
Initial working implementation of prisoner's visitor list. Needs a few more tests and potentially some style tweaks.

### DIFF
--- a/server/errorHandler.test.ts
+++ b/server/errorHandler.test.ts
@@ -27,7 +27,7 @@ describe('GET 404', () => {
   it('should render content without stack in production mode', () => {
     const systemToken = async (user: string): Promise<string> => `${user}-token-1`
 
-    return request(appWithAllRoutes(null, null, systemToken, true))
+    return request(appWithAllRoutes(null, null, null, systemToken, true))
       .get('/unknown')
       .expect(404)
       .expect('Content-Type', /html/)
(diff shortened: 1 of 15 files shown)


server/errorHandler.test.ts
```diff
@@ -27,7 +27,7 @@ describe('GET 404', () => {
   it('should render content without stack in production mode', () => {
     const systemToken = async (user: string): Promise<string> => `${user}-token-1`
 
-    return request(appWithAllRoutes(null, null, systemToken, true))
+    return request(appWithAllRoutes(null, null, null, systemToken, true))
```
psoleckimoj, line 30: Thinking at some point we change this to an object to make it easier to add stuff 🙂

server/utils/nunjucksSetup.ts
```diff
@@ -45,4 +46,8 @@ export default function nunjucksSetup(app: express.Express, path: pathModule.Pla
     const array = fullName.split(' ')
     return `${array[0][0]}. ${array.reverse()[0]}`
   })
+
+  njkEnv.addFilter('formatDate', (dateToFormat, dateFormat = 'd MMMM yyyy') => {
```
psoleckimoj, line 50: We could add a few filters for different types of dates and keep the generic one too? Makes it easier to see and update places where it's used maybe?

psoleckimoj (approved):
LGTM, only one comment which can be sorted if we want to.

PR #60, opened 2022-02-08
Add choose visit date and time slot

psoleckimoj:


server/@types/bapv.d.ts
```diff
@@ -143,3 +143,23 @@ export type Contact = {
   addresses: Address[]
   commentText: string
 }
+
+// Visit slots, for representing data derived from VisitSessions
```
tpmcgowan, line 147: Happy to discuss naming of these - 'slots' vs 'sessions', etc! (Might try and order this file better too once this PRs in...)

server/services/visitSessionsService.ts
```diff
@@ -0,0 +1,58 @@
...
+        slotsForDay.slots.morning.push(newSlot)
+      } else {
+        slotsForDay.slots.afternoon.push(newSlot)
+      }
+      return slotList
+    }, {})
+  }
+
+  // include minutes only if not on the hour; e.g. 10am / 10:30am
+  private getFormattedTime(time: Date): string {
```
tpmcgowan, line 55: This works OK for now, but thinking it probably makes sense to move this to utils or add as a Nunjucks filter (and hence just pass timestamps to template).

server/@types/bapv.d.ts
psoleckimoj: Can go with whatever we use, one for the team 🙂

tpmcgowan (approved):
👍

PR #61, opened 2022-02-08
Fix options not deselecting when nothing selected

psoleckimoj:


server/routes/visitors.ts
```diff
@@ -27,40 +27,40 @@ export default function routes(router: Router, prisonerVisitorsService: Prisoner
...
+      }
+
+      if (selected.length > 3) {
+        throw new Error('Select no more than 3 visitors with a maximum of 2 adults')
+      }
+
+      const adults = req.session.visitorList
+        .filter((visitor: VisitorListItem) => selected.includes(visitor.personId.toString()))
+        .reduce((count: number, visitor: VisitorListItem) => {
+          return visitor.adult ? count + 1 : count
```
tpmcgowan, line 51: Could potentially change to:
`return visitor.adult ?? true ? count + 1 : count`
which I think would stop visitors with no date of birth (so `adult` property is `undefined`) causing issues by effectively counting them as adults - but probably need confirmation on what it *should* do

tpmcgowan (dismissed):
Tested and works OK for me! (Just a comment about potentially how to handle those with no date of birth...)

server/routes/visitors.ts
psoleckimoj: 👍

tpmcgowan (approved):
LGTM!

PR #63, opened 2022-02-09
VB-326 Get visit schedule

tpmcgowan:


server/@types/bapv.d.ts
```diff
@@ -56,28 +56,6 @@ export type PastVisitItem = [
   }
 ]
 
-type VisitVisitor = {
```
tpmcgowan, line 59: I removed these in favour of auto-generated ones from the Visit Scheduler API.

psoleckimoj (approved):
LGTM

PR #69, opened 2022-02-17
Refactor Visit Slots and and add Nunjucks filter

tpmcgowan:
Keep hold of VisitSession timestamps as they will be needed to make the session booking.

Move time formatting to a Nunjucks filter so it can be re-used.

### DIFF
--- a/server/utils/nunjucksSetup.test.ts
+++ b/server/utils/nunjucksSetup.test.ts
@@ -1,0 +1,123 @@
+import cheerio from 'cheerio'
+import nunjucks, { Template } from 'nunjucks'
+import { registerNunjucks } from './nunjucksSetup'
+
+describe('Nunjucks Filters', () => {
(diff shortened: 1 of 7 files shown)


server/utils/nunjucksSetup.test.ts
```diff
@@ -0,0 +1,123 @@
...
+import nunjucks, { Template } from 'nunjucks'
+import { registerNunjucks } from './nunjucksSetup'
+
+describe('Nunjucks Filters', () => {
+  let compiledTemplate: Template
+  let viewContext: Record<string, unknown>
+
+  const njkEnv = registerNunjucks()
+
+  describe('initialiseName', () => {
```
tpmcgowan, line 11: Got these tests for this filter in the template app from another project; figured we may as well have it...

psoleckimoj (approved):
LGTM!

PR #71, opened 2022-02-17
VB-326 Add offenderNo to session to fix 404s

tpmcgowan:
`offenderNo` wasn't coming through on POST requests (following failed validation) to get visitor list. Added to the session on the initial GET so it's available. Added test coverage for this.

### DIFF
--- a/server/@types/express/index.d.ts
+++ b/server/@types/express/index.d.ts
@@ -1,4 +1,4 @@
-import { Contact, VisitorListItem } from '../bapv'
+import { VisitorListItem } from '../bapv'
 
 export default {}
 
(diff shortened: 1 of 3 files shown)


server/@types/express/index.d.ts
```diff
@@ -8,7 +8,7 @@ declare module 'express-session' {
     returnTo: string
     nowInMinutes: number
     prisonerName: string
-    contacts: Contact[]
```
tpmcgowan, line 11: Not currently storing the raw `Contacts[]` data in session so removed for now.

psoleckimoj (approved):
LGTM

PR #72, opened 2022-02-18
Automatically remove whitespace in Nunjucks

tpmcgowan:
These config options are giving much cleaner (and smaller!) HTML output on all pages. For date/time picker specifically there were some big chunks of whitespace caused by loops, have tweaked these in the template.


psoleckimoj (approved):
LGTM!

PR #77, opened 2022-02-24
VB-328: Additional support needs

tpmcgowan:


server/routes/visitors.ts
```diff
@@ -113,8 +113,8 @@ export default function routes(
 
       return true
     }),
-    query('timeOfDay').custom((value: string) => (!['morning', 'afternoon'].includes(value) ? '' : value)),
-    query('dayOfTheWeek').custom((value: string) =>
+    query('timeOfDay').customSanitizer((value: string) => (!['morning', 'afternoon'].includes(value) ? '' : value)),
```
tpmcgowan, line 116: This wasn't working as a validator (not returning true/false correctly, I think), but replaced as a sanitiser so it actually sets to defaults if any disallowed values passed.

server/@types/express/index.d.ts
```diff
@@ -11,6 +11,9 @@ declare module 'express-session' {
     offenderNo: string
     visitorList: VisitorListItem[]
     adultVisitors: VisitorListItem[]
+    slotsList: VisitSlotList
```
tpmcgowan, line 14: Added these to session so list of slots and filter settings can be saved on GET request to be re-rendered in POST route should validation fail. For `timeOfDay` and `dayOfWeek`, I did consider just keeping these in the URL params but that would have meant duplication of the validation....

server/routes/visitors.ts
```diff
@@ -137,6 +141,70 @@ export default function routes(
...
+          dayOfTheWeek: req.session.dayOfTheWeek,
+        })
+      }
+      return res.redirect(`/visit/additional-support/${req.params.offenderNo}`)
+    }
+  )
+
+  router.get(
+    '/additional-support/:offenderNo',
+    param('offenderNo').custom((value: string) => {
```
tpmcgowan, line 190: I added this check on `offenderNo` for consistency with the other routes but have realised it doesn't do anything here (and in some other places) because it's not showing errors on `GET` requests.

Have left for time being as thought it worth discussing whether we need to keep the `offenderNo` in the URL all the way through? It's not actually doing anything on some screens; i.e. you could put a different (but valid) one - and be confused! Thinking it might be better to keep it there but use the validation to check it matches the one in the session? (And if it didn't, probably send you back to the beginning as something's gone wrong....)

server/routes/visitors.ts
psoleckimoj: It was returning either an empty string or the value which is incorrect I know realise, it should return true if there is no error rather than any sort of value.

server/routes/visitors.ts
psoleckimoj: I'd already thought about that tbh, after the initial selection I don't think it's required. It should go in the session 🙂

server/routes/visitors.test.ts
```diff
@@ -277,3 +277,116 @@ describe('POST /visit/select-visitors/A1234BC', () => {
     })
   })
 })
+
+describe.only('POST /visit/additional-support/:offenderNo', () => {
```
psoleckimoj, line 281: You've got a `.only` here.

server/routes/visitors.ts
```diff
@@ -137,6 +141,127 @@ export default function routes(
     }
   )
 
+  router.post(
+    '/select-date-and-time/:offenderNo',
+    body('visit-date-and-time').custom((value: string, { req }) => {
+      const slotsList = req.session.slotsList as VisitSlotList
+
+      // check selected slot is in the list that was shown and has available tables
+      const selectedSlot: VisitSlot = Object.values(slotsList)
```
psoleckimoj, line 150: Do you need this with the `.flat()` ? Can't you just do `slotsList.reduce....` ? I'm probably missing something, but `reduce` will give you the value rather than the key anyway?

psoleckimoj: Also would `.filter` work rather than `reduce`?

PR #78, opened 2022-02-28
VB-332: Initial stab at type for visit session data

psoleckimoj:


server/@types/bapv.d.ts
```diff
@@ -170,3 +170,24 @@ export type FormError = {
   param: string
   location: string
 }
+
+export type visitSessionData = {
```
tpmcgowan, line 174: Looks good 👍. Will probably need to add something like:
```
export type additionalSupportRequest = {
   value: string
   label: string
}
```

and then `additionalSupport: additionalSupportRequest[]` within `visitSessionData`? I'm thinking this would be populated, for example, like:
```
additionalSuuport: [
  { value: "ramp", label: "Wheelchair ramp" },
  { value: "other", label: "Some other requirement..." },
  // ...
]
```

Also, wondering whether we'd need to mark the 'sections' (`visit`, `mainContact`, etc) as optional to prevent TypeScript errors? Not sure...🤔

psoleckimoj: Sorry was just a high level overview, but needs breaking down. I didn't want to store the additional Support values in the session as they're in a config/enums file so only need the key to be able to display the info or to retrieve it again? Or at least that was my thinking?

tpmcgowan: Yeah, you're right - for most of the additional support options we could just store the keys in the session and that would obviously make a lot more sense. Complication though is the 'other' option where they can enter free text for the value. But, I guess that can be solved by just having two fields; one for the keys of the default additional request types (e.g. 'ramp') and another just for any free text 'other' request.

PR #80, opened 2022-03-01
VB-332: Create initial session

psoleckimoj:


server/@types/express/index.d.ts
```diff
@@ -17,6 +17,7 @@ declare module 'express-session' {
     contact: string
     phoneNumber: string
     someoneElseName: string
+    visitSessionData: VisitSessionData
```
psoleckimoj, line 20: I'll remove the other fields once we've fully moved over to this object type 👍

tpmcgowan (approved):
LGTM!

PR #86, opened 2022-03-03
Prevent form re-submission warnings and pre-populate forms with session data (additional support)

tpmcgowan:
Also pre-populates form with session data if available. Some tests skipped while this work in progress.

### DIFF
--- a/server/routes/testutils/appSetup.ts
+++ b/server/routes/testutils/appSetup.ts
@@ -29,6 +29,8 @@ const user = {
   displayName: 'John Smith',
 }
 
+export const flashProvider = jest.fn()
+
 class MockUserService extends UserService {
   constructor() {
     super(undefined)
@@ -70,6 +72,7 @@ function appSetup(
   app.use((req, res, next) => {
     res.locals = {}
     res.locals.user = req.user
+    req.flash = flashProvider
     req.session = {
       ...sessionData,
       regenerate: jest.fn(),
@@ -116,7 +119,7 @@ function appSetup(
   return app
 }
 
-export default function appWithAllRoutes(
+export function appWithAllRoutes(
   prisonerSearchServiceOverride?: PrisonerSearchService,
   prisonerProfileServiceOverride?: PrisonerProfileService,
   prisonerVisitorsServiceOverride?: PrisonerVisitorsService,
--- a/server/routes/visitors.test.ts
+++ b/server/routes/visitors.test.ts
@@ -289,7 +289,7 @@ describe('GET /visit/select-main-contact/A1234BC', () => {
   })
 })
 
-describe('POST /visit/additional-support/:offenderNo', () => {
+describe.skip('POST /visit/additional-support/:offenderNo', () => {
   it('should show error if additional support question not answered', () => {
     const req = request(app).post('/visit/additional-support/A1234BC')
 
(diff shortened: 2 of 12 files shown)


server/routes/visitors.ts
```diff
@@ -254,22 +261,20 @@ export default function routes(
...
-          additionalSupportRequired: req.body.additionalSupportRequired,
-          additionalSupport: req.body.additionalSupport,
-          otherSupportDetails: req.body.otherSupportDetails,
-        })
+        req.flash('errors', errors.array() as [])
+        req.flash('formValues', req.body)
+        return res.redirect(req.originalUrl)
       }
 
       let selectedSupport
```
psoleckimoj, line 269: Could rearrange this to make it a `const`, the default value would be what you're doing in the else?

server/routes/testutils/appSetup.ts
```diff
@@ -29,6 +29,8 @@ const user = {
   displayName: 'John Smith',
 }
 
+export const flashProvider = jest.fn()
```
psoleckimoj, line 32: Doesn't seem to be imported anywhere?

tpmcgowan: Gets used in `server/routes/visitors.test.ts`

psoleckimoj: Sorry, there were so many changes it didn't show and I assume it was a `package-lock.json` 😂

server/routes/visitors.test.ts
```diff
@@ -1,15 +1,19 @@
...
+import { VisitorListItem, VisitSessionData } from '../@types/bapv'
 import PrisonerVisitorsService from '../services/prisonerVisitorsService'
-import appWithAllRoutes from './testutils/appSetup'
+import { appWithAllRoutes, flashProvider } from './testutils/appSetup'
 
 let app: Express
+let sessionApp: Express
 let prisonerVisitorsService: PrisonerVisitorsService
-let systemToken
+let systemToken: (user: string) => Promise<string>
```
psoleckimoj, line 12: Actually occurred to me while doing my test updates that this can just be set to a `const` as the value it is on line 30 as it never changes.

server/routes/visitors.test.ts
```diff
@@ -279,125 +286,318 @@ describe('POST /visit/select-visitors/A1234BC', () => {
...
-    })
+  it('should render the additional support page with no options selected', () => {
+    sessionApp = appWithAllRoutes(null, null, null, null, systemToken, false, { visitSessionData } as SessionData)
+
+    return request(sessionApp)
+      .get('/visit/additional-support/A1234BC')
+      .expect(200)
+      .expect('Content-Type', /html/)
+      .expect(res => {
+        const $ = cheerio.load(res.text)
```
psoleckimoj, line 309: This looks a great way to do this 🙂

psoleckimoj (approved):
LGTM

server/routes/visitors.test.ts
tpmcgowan: Feel I should declare creative inspiration from `hmpps-welcom-people-into-prison-ui` 😉!

server/routes/visitors.test.ts
tpmcgowan: 👍 will do this as part of a subsequent PR

PR #89, opened 2022-03-04
VB-332: Fix session tests

psoleckimoj:
Fix the session related tests by creating a session object in the test app.


tpmcgowan (approved):
LGTM!

PR #90, opened 2022-03-05
Sort timeout correctly for HttpAgent

psoleckimoj:
Copied from https://github.com/ministryofjustice/hmpps-template-typescript/pull/69/files

Thought it might be the cause of some of our issues.


tpmcgowan (approved):
LGTM

PR #92, opened 2022-03-08
VB-332: Check session in journey

psoleckimoj:


server/routes/visitors.ts
```diff
@@ -18,6 +17,55 @@ const getSelectedSlot = (slotsList: VisitSlotList, selectedSlot: string): VisitS
...
+  visitData,
+  res,
+}: {
+  stage: number
+  visitData: VisitSessionData
+  res: Response
+  // eslint-disable-next-line consistent-return
+}): void => {
+  if (!visitData) {
+    return res.redirect('/search/')
```
tpmcgowan, line 31: Not necessary for now, but I think something before the redirects like `req.flash('errors', [{ msg: 'Session expiry/error message...' }]` would be picked up if there was the errorSummary component included on the target page (and it had `{ errors: req.flash('errors') }`) in the context passed to view.

tpmcgowan (approved):
LGTM - only suggestion would be adding a few tests to poke in the 'wrong' session data and confirm it redirects. I don't _think_ that's covered but tbh I was getting a bit lost in that `visitors.test.ts` file 😵‍💫!

server/routes/visitors.ts
psoleckimoj: This was an MVP as we discussed so we can always alter it. Basically if there's an error with the session it'll just throw you back to the point it can do.

psoleckimoj:
@tpmcgowan definitely needs more tests! I want to do some tests around the check session function.

PR #101, opened 2022-03-10
VB-574: Select visitors - form fixes and pre-populate from session

tpmcgowan:


server/routes/visitors.test.ts
```diff
@@ -37,21 +37,16 @@ afterEach(() => {
...
+        dateOfBirth: '25 May 1988',
+        location: 'location place',
       },
-    } as SessionData)
-  })
+    }
 
-  it('should render the approved visitor list for offender number A1234BC', () => {
     returnData = {
       prisonerName: 'John Smith',
```
tpmcgowan, line 51: I noticed that we've got a bit of duplication with `prisonerName` coming from the visitor list service yet it is now stored in the session. Will tidy this up separately so as to change things in smaller chunks!

psoleckimoj (approved):
LGTM!

PR #103, opened 2022-03-11
VB-334: Add reserve visit after selecting date and time for slot

psoleckimoj:


server/data/restClient.ts
```diff
@@ -23,6 +23,14 @@ interface PostRequest {
   raw?: boolean
 }
 
+interface PutRequest {
```
psoleckimoj, line 26: Actually not required at this stage but forgot to remove, next step will be a PUT to update the reservation and complete.

server/@types/bapv.d.ts
```diff
@@ -191,4 +191,5 @@ export type VisitSessionData = {
     phoneNumber: string
     contactName?: string
   }
+  reservationId?: number
```
tpmcgowan, line 194: Do we also need to store the status (i.e. `RESERVED` or `BOOKED`)? E.g. `reservation { id: number, status: 'RESERVED' | 'BOOKED' }`. May well not need to; I suppose it depends what we need at the final stages...

server/routes/visitors.ts
```diff
@@ -177,6 +177,13 @@ export default function routes(
 
       req.session.visitSessionData.visit = getSelectedSlot(req.session.slotsList, req.body['visit-date-and-time'])
 
+      const reservationId = await visitSessionsService.reserveVisit({
```
tpmcgowan, line 180: Some cases I think we need to eventually handle here:
* reservation fails (e.g. slot now full). `try/catch` and then setting error message like "session full, try again..."? Theoretical at the moment as I think the back-end currently not checking capacity, etc!
* `reservationId` already set and user, e.g., goes back to this page from check details page:
  1. form re-posted with same slot - would duplicate booking be created? Could probably avoid by checking if chosen slot same as that already in session.
  2. form re-posted with new slot - looks like this would record the new `reservationId` and leave the previous reservation orphaned. Could either cancel existing reservation first or leave the orphaned one to 'time out' I the back-end? Probably one to discuss with Phil, Richard, etc.

tpmcgowan (approved):
LGTM - a few comments on things I think we need to consider but these could be done separately.

server/routes/visitors.ts
psoleckimoj: I've merged but will look at this next 👍

PR #105, opened 2022-03-14
VB-334: More checks for existing reservations

psoleckimoj:
- Rename `reserveVisit` to `createVisit` as it's more generic
- Add `updateVisit` for updating an existing visit when revisiting certain pages
- `updateVisit` will be used when updating the final reservation on the confirmation page


tpmcgowan (approved):
LGTM!

PR #106, opened 2022-03-14
VB-334: Add confirmation page

psoleckimoj:


server/views/pages/confirmation.njk
```diff
@@ -0,0 +1,174 @@
...
+            classes: "govuk-!-margin-top-3, govuk-!-margin-bottom-6",
+            text: "Manage prison visits"
+        }) }}
+    </div>
+</div>
+
+<div class="govuk-grid-row">
+    <div class="govuk-grid-column-full">
+        <h2 class="govuk-heading-m">Visit details</h2>
+        {{ govukSummaryList({
```
tpmcgowan, line 50: Does this end up pretty similar to the content on check your answers page - could perhaps have it as an include template?

server/routes/visitors.ts
```diff
@@ -413,5 +410,93 @@ export default function routes(
...
+    const { visitSessionData } = req.session
+    checkSession({
+      stage: 5,
+      visitSessionData,
+      res,
+    })
+
+    const { offenderNo } = req.session.visitSessionData.prisoner
+
+    const additionalSupport = visitSessionData.additionalSupport?.keys?.map(key => {
```
tpmcgowan, line 423: Could move this to utils function as same is used in GET route? Though perhaps not worth it at this stage as we'll probably need change additional support a bit once Phil's back-end stuff done...

tpmcgowan (approved):
LGTM!

server/views/pages/confirmation.njk
psoleckimoj: Irritatingly it's not the same, a few bits are, maybe I can pull out those sections which are the same.

server/routes/visitors.ts
psoleckimoj: I did wonder, let's wait for us to pull in Phil's changes and go from there.

server/views/pages/confirmation.njk
psoleckimoj: I'll do this after I've merged this.

PR #107, opened 2022-03-14
VB-574: Main contact - form fixes and pre-populate from session

tpmcgowan:


server/routes/visitors.ts
```diff
@@ -290,11 +295,22 @@ export default function routes(
       res,
     })
 
-    const { offenderNo } = req.session.visitSessionData.prisoner
+    const formValues = (req.flash('formValues')?.[0] as unknown as Record<string, string | string[]>) || {}
```
psoleckimoj, line 298: If this is used everywhere, maybe stick in a function?

server/routes/visitors.ts
```diff
@@ -290,11 +295,22 @@ export default function routes(
...
+
+    if (!Object.keys(formValues).length && req.session.visitSessionData.mainContact) {
```
psoleckimoj, line 300: Maybe assign `req.session.visitSessionData` to a variable?

psoleckimoj: ..to make it easier to read.

psoleckimoj (dismissed):
All looks good with a couple of minor comments. Also you'll have to explain why the data structure needed changing as I can't see the reason from the code 🤷‍♂️

psoleckimoj (approved):
LGTM!

server/routes/visitors.ts
tpmcgowan: Good suggestion - have done this 👍

server/routes/visitors.ts
tpmcgowan: Done - in fact as it was repeated a lot I got carried away and did it throughout this file and the utils file.

psoleckimoj (approved):
LGTM!

tpmcgowan:
@psoleckimoj about the data structure change, I'm assuming you were referring to e.g. `adultVisitors: VisitorListItem[]` becoming `adultVisitors: { adults: VisitorListItem[] }`? This was because when I tried to pass `adultVisitors` via `appWithAllRoutes(...)` I found I couldn't test what had happened as it got overwritten in the code. Passing an object with this value inside seemed to let me inspect it afterwards to see what had happened.

PR #109, opened 2022-03-15
VB-474: Add offender restrictions to Add Visitors

psoleckimoj:


server/views/pages/visitors.njk
```diff
@@ -81,6 +81,53 @@
...
+              {
+                text: restriction.comment,
+                classes: "test-restrictions-comment" + loop.index
+              },
+              {
+                text: restriction.startDate | formatDate,
+                classes: "test-restrictions-start-date" + loop.index
+              },
+              {
+                text: restriction.expiryDate | formatDate,
```
tpmcgowan, line 104: I think you might need to add 'Not entered' if there's no expiry date?(Looking at the Prison API, neither start nor end date are required fields so perhaps both could actually be missing?) Could add a second test restriction in `visitors.test.ts` with no date(s) to cover this.

tpmcgowan (approved):
LGTM!

PR #110, opened 2022-03-15
Fix update/create visit logic on select date and time page

tpmcgowan:
* Fix create vs update visit logic
* Add logging for create/update visit operations
* Test coverage for calls to create/update visit session service (only in POST /visit/select-date-and-time)

### DIFF
--- a/server/routes/visitors.test.ts
+++ b/server/routes/visitors.test.ts
@@ -697,14 +697,19 @@ describe('/visit/select-date-and-time', () => {
   })
 
   describe('POST /visit/select-date-and-time', () => {
+    const visitSessionsService = new VisitSessionsService(null, systemToken) as jest.Mocked<VisitSessionsService>
+
     beforeEach(() => {
-      sessionApp = appWithAllRoutes(null, null, null, null, systemToken, false, {
+      visitSessionsService.createVisit = jest.fn().mockResolvedValue('2a-bc-3d-ef')
+      visitSessionsService.updateVisit = jest.fn()
+
+      sessionApp = appWithAllRoutes(null, null, null, visitSessionsService, systemToken, false, {
         slotsList,
         visitSessionData,
       } as SessionData)
     })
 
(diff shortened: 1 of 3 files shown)


server/routes/visitors.test.ts
```diff
@@ -697,14 +697,19 @@ describe('/visit/select-date-and-time', () => {
   })
 
   describe('POST /visit/select-date-and-time', () => {
+    const visitSessionsService = new VisitSessionsService(null, systemToken) as jest.Mocked<VisitSessionsService>
+
     beforeEach(() => {
-      sessionApp = appWithAllRoutes(null, null, null, null, systemToken, false, {
+      visitSessionsService.createVisit = jest.fn().mockResolvedValue('2a-bc-3d-ef')
```
psoleckimoj, line 703: Do we need to do this here if we're not reseting them on an `afterEach` ?

server/services/visitSessionsService.ts
```diff
@@ -70,6 +71,7 @@ export default class VisitSessionsService {
     const visitSchedulerApiClient = this.visitSchedulerApiClientBuilder(token)
 
     const reservation = await visitSchedulerApiClient.createVisit(visitData)
+    logger.info(`Created visit ${reservation?.id} for offender ${visitData.prisoner.offenderNo}`)
```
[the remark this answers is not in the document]
psoleckimoj: Good idea!

psoleckimoj (approved):
LGTM!

server/routes/visitors.test.ts
tpmcgowan: Just tried that but it didn't work - they're getting reset on an `afterEach()` right at the top of the file. One to consider when we split up this visits stuff!

psoleckimoj: I think it's all to do with scope, but yeah we can address it all when we split it up 👍

PR #111, opened 2022-03-16
Add eslint no-only-tests rule

tpmcgowan:
This was added to the template project recently; been meaning to add it here since I accidentally committed an `.only` test a while ago...


psoleckimoj (approved):
LGTM!

PR #112, opened 2022-03-16
Main contact choice: use ID for radio value

tpmcgowan:
Previously value attribute of the radio buttons to choose the main contact was `firstname_lastname`, but this wouldn't cater for two visitors with the same name. Replaced with the `personId` value.


psoleckimoj (approved):
LGTM!

PR #113, opened 2022-03-16
Date/time page: ensure accordion is expanded if slot inside selected

tpmcgowan:
On select date/time page, if there is a selected visit slot, set the `expanded` attribute for that accordion section to ensure it is visible.

Found in testing that if you've been opening/closing the sections and going back & forth you could end up viewing this page with the selected slot hidden within a collapsed accordion section.


psoleckimoj (approved):
Nice! LGTM!

PR #116, opened 2022-03-17
Refactor date/time slot template to use macro

tpmcgowan:
Use Nunjucks macro to remove duplication when generating morning / afternoon slot radio buttons.


psoleckimoj (approved):
LGTM!

PR #118, opened 2022-03-17
Tidy prisonerVisitorService:  only return visitor list

tpmcgowan:
This was returning prisoner name too, requiring another API call. No longer needed now that prisoner name is stored in session; meaning a lot of unnecessary code can be removed.


server/@types/bapv.d.ts
```diff
@@ -90,9 +90,9 @@ export type Restriction = {
   restrictionType: string
   restrictionTypeDescription: string
   startDate: string
-  expiryDate: string
+  expiryDate?: string
```
tpmcgowan, line 93: Made these optional to match the API schema - and to placate TypeScript when I typed the test data.

psoleckimoj (approved):
LGTM!

PR #121, opened 2022-03-20
Fix docker caching as per template app pull 74

psoleckimoj:
Update re

https://github.com/ministryofjustice/hmpps-template-typescript/pull/74/files


tpmcgowan (approved):
LGTM

PR #123, opened 2022-03-24
VB-593: Refactor additional support following API changes

tpmcgowan:
* Update types from visit scheduler
* Use list of available support types retrieved from visit scheduler (and then stored in session) rather than hard-coded list
* Update visit creation/update to match changes to the visit scheduler API
* Update and add to tests to reflect these changes


server/@types/bapv.d.ts
```diff
@@ -134,11 +135,7 @@ export type VisitSessionData = {
   }
   visit?: VisitSlot
   visitors?: VisitorListItem[]
-  additionalSupport?: {
-    required: boolean
-    keys?: string[]
-    other?: string
-  }
+  visitorSupport?: VisitorSupport[]
```
tpmcgowan, line 138: Changed the format of the data we store in the session to match what gets saved to the visit scheduler to avoid converting back/forth. So, using array of `VisitorSupport` items and just checking length rather than separate `required` value.

server/@types/express/index.d.ts
```diff
@@ -7,6 +8,7 @@ declare module 'express-session' {
   interface SessionData {
     returnTo: string
     nowInMinutes: number
+    availableSupportTypes: SupportType[]
```
tpmcgowan, line 11: On first request, the list of available `SupportType` options (names and descriptions) is retrieved from the visit scheduler and stored here.

psoleckimoj (approved):
LGTM!

PR #133, opened 2022-04-12
VB-639: Add search by visit reference

psoleckimoj:


server/routes/search.ts
```diff
@@ -65,24 +68,75 @@ export default function routes(router: Router, prisonerSearchService: PrisonerSe
...
         },
       })
     )
   })
 
+  router.get('/visit/results', async (req, res) => {
+    const { searchBlock1, searchBlock2, searchBlock3, searchBlock4 } = req.query
+    const search = `${searchBlock1}-${searchBlock2}-${searchBlock3}-${searchBlock4}`
+
+    const currentPage = '1'
```
psoleckimoj, line 97: Much is hardwired because it's not a search, it will only ever return a single result.

tpmcgowan (approved):
LGTM!

PR #137, opened 2022-04-22
Simplify tables: style table rather than each cell

tpmcgowan:
Simplify markup for tables by adding the extra spacing to the whole table (using class `bap-table`) rather than to every cell (using `bapv-table_cell`).


psoleckimoj (approved):
LGTM

PR #140, opened 2022-04-25
VB-683 use visitStatus filter

tpmcgowan:
* update the types from the Visit Scheduler
* add the `visitStatus` filter (hard-coded to `BOOKED`) to `GET /visits` requests so that cancelled or reserved visits don't show in listings (e.g. on prisoner profile tabs or prisoner visits listing)


psoleckimoj (approved):
LGTM!

PR #142, opened 2022-04-26
VB-680: Visit booking summary (initial implementation)

tpmcgowan:
Initial implementation of the visit booking summary page. Route in place, service and template showing most of the data.

Need to get the full details showing in the visitor list and additional support request information but this will need a bit of refactoring to avoid duplication - so getting this in now to keep the changes smaller.


server/services/visitSessionsService.ts
```diff
@@ -205,14 +209,46 @@ export default class VisitSessionsService {
...
+    reference,
+  }: {
+    username: string
+    reference: string
+  }): Promise<{ visit: Visit; visitors: Contact[] }> {
+    const token = await this.systemToken(username)
+    const visitSchedulerApiClient = this.visitSchedulerApiClientBuilder(token)
+    const prisonerContactRegistryApiClient = this.prisonerContactRegistryApiClientBuilder(token)
+
+    logger.info(`Get visit ${reference}`)
```
psoleckimoj, line 231: Loggers might be better in the client? That way you'd get the log on every call? Pretty sure the REST client logs every request anyway but better in client rather than having to do this everywhere it's called?

server/services/visitSessionsService.ts
```diff
@@ -205,14 +209,46 @@ export default class VisitSessionsService {
...
+    logger.info(`Get contacts for ${visit.prisonerId}`)
+    const contacts = await prisonerContactRegistryApiClient.getPrisonerSocialContacts(visit.prisonerId)
+
+    const visitorIds: number[] = visit.visitors.reduce((acc, visitor) => {
+      acc.push(visitor.nomisPersonId)
+      return acc
+    }, [])
+
+    const visitors: Contact[] = contacts.reduce((acc, contact) => {
+      if (visitorIds.includes(contact.personId)) {
```
psoleckimoj, line 243: Could possibly to a `.filter()` on this rather than the `if` statement?

psoleckimoj (approved):
LGTM, just a couple of points but won't stop merging.

server/services/visitSessionsService.ts
tpmcgowan: Some stray debugging - will remove!

server/services/visitSessionsService.ts
tpmcgowan: Good suggestion, will give that a try 👍

PR #151, opened 2022-04-29
VB-570: Add test for prisoner profile page

psoleckimoj:
Lots of changes:

- Updated types to allow data attributes for testing
- Added integration tests for prisoner profile
- Refactored test stubs to allow data to be passed in
- Lots more!!


server/services/prisonerProfileService.ts
```diff
@@ -152,31 +187,55 @@ export default class PrisonerProfileService {
...
-          { text: `${properCase(visit.visitStatus)}` },
         ] as PastVisitItem
       })
     )
 
     return visitsForDisplay
   }
 
-  private async getPrisonerSocialContacts(contacts: Contact[], contactIds: number[]): Promise<string[]> {
+  private getPrisonerSocialContacts(contacts: Contact[], visitors: Visitor[]): string[] {
```
psoleckimoj, line 232: Not strictly needed but spotted it whilst doing other bits, removes more duplication.

server/utils/utils.ts
```diff
@@ -109,14 +109,18 @@ export const convertToTitleCase = (sentence: string): string =>
 export const visitDateAndTime = ({
   startTimestamp,
   endTimestamp,
+  isIntTest = false,
```
psoleckimoj, line 112: Made this an optional, allows for the integration tests to return a different format for checking as cypress removes HTML tags.

server/services/prisonerProfileService.ts
psoleckimoj: Also realised it wasn't actually an async function.

tpmcgowan (approved):
LGTM!

PR #169, opened 2022-05-09
VB-741: Refactor to use POST route to start booking journey

tpmcgowan:
Current Prisoner Profile page behaviour is to:
* Render 'Book a prison visit' button as a link to continue the journey (on to Select Visitors)
* Populate `visitSessionData` with the requested prisoner on load

In preparation for VB-741 (adding VO override checkbox), this changes the page to:
* use the button to submit to a new POST route that redirects to the next step in the journey (and will be used to validate the checkbox)
* populates `visitSessionData` in the POST route so that the session is only 'started' when the Book button is pressed rather than just by viewing a profile

### DIFF
--- a/server/routes/prisoner.test.ts
+++ b/server/routes/prisoner.test.ts
@@ -1,7 +1,8 @@
 import type { Express } from 'express'
 import request from 'supertest'
 import * as cheerio from 'cheerio'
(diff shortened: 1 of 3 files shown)


psoleckimoj (approved):
LGTM!

PR #171, opened 2022-05-09
Correct DNS names for this and dependent services

rj-adams:
This will break affected environments until all are in sync so we need to merge at appropriate time


psoleckimoj (approved):
LGTM assuming the backend changed for visit scheduler and contact registry go in at the same time or this will break the front end until it's done :)

PR #172, opened 2022-05-11
Add VO override checkbox to Prisoner Profile page

tpmcgowan:


server/routes/prisoner.test.ts
```diff
@@ -53,8 +59,10 @@ afterEach(() => {
 })
 
 describe('GET /prisoner/A1234BC', () => {
-  it('should render the prisoner profile page for offender number A1234BC', () => {
-    const returnData: PrisonerProfile = {
+  let prisonerProfile: PrisonerProfile
```
tpmcgowan, line 62: It looks like there's more going on in this file than is really the case. Basically, I've added some new tests (mainly for the the new POST route), but also 'converted' some existing tests to use `cheerio` and shuffled test data a little to avoid duplication.

psoleckimoj (approved):
LGTM!

PR #178, opened 2022-05-16
VB-752: Add initial view by visits layout

psoleckimoj:


server/config.ts
```diff
@@ -4,7 +4,11 @@ const production = process.env.NODE_ENV === 'production'
 
 function get<T>(name: string, fallback: T, options = { requireInProduction: false }): T | string {
   if (process.env[name]) {
-    return process.env[name]
+    if (production) {
+      return process.env[name]
+    }
+
+    return process.env[name].replace(/\\/g, '')
```
tpmcgowan, line 11: Is this the fix for `.env` issues you had?

tpmcgowan (approved):
LGTM!

server/config.ts
psoleckimoj: Yep, I left production as is, although maybe I need to work out when we're running locally?

tpmcgowan: 👍 perhaps you could just have something else in the `.env` file, like `MESSY_TOKEN=true`!

PR #180, opened 2022-05-16
Format contact with middle name correctly

tpmcgowan:
If a custom main contact was entered with a middle name (e.g. `Firstname Middlename Lastname`) this was rendered in the booking summary incorrectly as `Middlename, Firstname`. This fix displays it as `Lastname, Firstname`.


psoleckimoj (approved):
LGTM!

PR #181, opened 2022-05-16
VB-794: Update to support new visit Outcome mappings

tpmcgowan:
Update to latest visit scheduler types and update code and tests to support the new structure for visit outcomes.


psoleckimoj (approved):
LGTM!

PR #185, opened 2022-05-19
Create slot time side menu

psoleckimoj:


server/routes/bookAVisit.test.ts
```diff
@@ -151,7 +151,7 @@ describe('GET /book-a-visit/select-visitors', () => {
       },
     }
 
-    prisonerVisitorsService.getVisitors.mockResolvedValue(returnData)
+    prisonerVisitorsService.getApprovedVisitors.mockResolvedValue(returnData)
```
psoleckimoj, line 154: Renamed this to make it match what it actually does. Could've added an arg to the function but it's obvious what it does when the function is named like this.

server/@types/bapv.d.ts
```diff
@@ -176,7 +176,7 @@ export type VisitSessionData = {
     location: string
   }
   visit?: VisitSlot
-  visitRestriction?: 'OPEN' | 'CLOSED'
+  visitRestriction?: 'OPEN' | 'CLOSED' | 'UNKNOWN'
```
tpmcgowan, line 179: I guess the `UNKNOWN` is needed to show migrated visits that may not have a type set?

server/services/prisonerVisitorsService.ts
```diff
@@ -25,4 +25,18 @@ export default class PrisonerVisitorsService {
 
     return visitorList
   }
+
+  async getAllVisitors(offenderNo: string, username: string): Promise<VisitorListItem[]> {
```
tpmcgowan, line 29: Might be missing something, but is this new `getAllVisitors()` actually used?

server/@types/bapv.d.ts
psoleckimoj: It was moaning at me that it was in the pulled in types.

PR #188, opened 2022-05-19
VB-810: Use visitRestriction (open/closed) in session and to filter available slots

tpmcgowan:
Set `visitRestriction` (defaulting to `OPEN`) when `visitSessionData` is first created. Use this value throughout journey (instead of hardcoded 'Open') and use it to filter available date/time slots.


psoleckimoj (approved):
LGTM!

PR #189, opened 2022-05-20
VB-806 Add UI roles to client in HMPPS Auth db dev scripts

AledWEvans:
This is the docker element to this, 95% working, prisoner-offender-search
has problems with elasticsearch


docker-compose-local.yml
```diff
@@ -0,0 +1,112 @@
...
+    restart: always
+    ports:
+      - "5432:5432"
+    environment:
+      - POSTGRES_PASSWORD=visit_scheduler
+      - POSTGRES_USER=visit_scheduler
+      - POSTGRES_DB=visit_scheduler
+
+networks:
+  hmpps:
```
psoleckimoj, line 112: Needs a newline.

docker-compose-local.yml
```diff
@@ -0,0 +1,112 @@
+version: '3.1'
```
psoleckimoj, line 1: Only comment is that this could all go in `docker-compose.yml` and you can just start the services you require by naming them in `docker-compose up`. That's all we do, usual for me is `docker-compose up redis-bapv` as that's all I need 🙂

psoleckimoj (approved):
LGTM!

PR #191, opened 2022-05-22
VB-826: Clear session after booking

tpmcgowan:
Better handle session data and ensure it it cleared out to prevent existing bookings being amended rather than creating new ones.
* keeps `visitStatus` in session and checks this and visit reference added to session check middleware
* session date cleared at start/end of booking journey
* updated tests

### DIFF
--- a/server/routes/bookAVisit.ts
+++ b/server/routes/bookAVisit.ts
@@ -5,7 +5,7 @@ import sessionCheckMiddleware from '../middleware/sessionCheckMiddleware'
 import PrisonerVisitorsService from '../services/prisonerVisitorsService'
 import PrisonerProfileService from '../services/prisonerProfileService'
 import VisitSessionsService from '../services/visitSessionsService'
-import { getFlashFormValues, getSelectedSlot, getSupportTypeDescriptions } from './visitorUtils'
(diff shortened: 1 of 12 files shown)


server/routes/bookAVisit.ts
```diff
@@ -392,33 +393,14 @@ export default function routes(
       visitSessionData.visitorSupport
     )
 
-    if (!req.session.visitSessionData.visitReference) {
```
tpmcgowan, line 395: Took this block out as it wouldn't get hit now that we have the session check middleware checking before it gets here

server/routes/prisoner.ts
```diff
@@ -48,7 +49,8 @@ export default function routes(
       }
     }
 
-    const visitSessionData: VisitSessionData = req.session.visitSessionData || { prisoner: undefined }
+    clearSession(req)
+    const visitSessionData: VisitSessionData = req.session.visitSessionData ?? { prisoner: undefined }
```
tpmcgowan, line 53: `req.session.visitData` shouldn't actually be present here after the session is cleared; this check is only there to enable `visitSessionData` to be passed through in testing when `clearSession` is mocked out. Couldn't think of a way around this....

psoleckimoj (approved):
LGTM!

PR #194, opened 2022-05-24
VB-748: Add tabs to visit search page

psoleckimoj:
`startDate` globally renamed as `selectedDate` for this page since `startDate` is unclear when you are talking about the tabs.


tpmcgowan (approved):
LGTM!

PR #199, opened 2022-05-25
VB-633: Amend add visitors page

psoleckimoj:
Add visitor fixes:

- When the prisoner has no active restrictions, do not display the restrictions tile
- Reduce the spacing between ‘prisoner name’ and their name to be consistent with next page
- Change placement of the rules text to be below approved visitor list title


tpmcgowan (approved):
LGTM

PR #200, opened 2022-05-25
VB-633: Update date picker as per ticket

psoleckimoj:
Date/time picker:

- Reduce the spacing between lines in the summary to be consistent with design and previous page
- Is there any way to reduce the spacing between elements in the accordion pattern? There is a lot of whitespace which makes the page feel long. This pattern probably needs to be more thoroughly tested (and accessibility tested)


tpmcgowan (approved):
LGTM

PR #201, opened 2022-05-25
VB-633: Amends to prisoner summary

psoleckimoj:
Prisoner summary fixes: 

- Update alerts and VOs/PVOs layout (apply to all the prisoner bio stuff - they should all be on the same line as the title, not underneath)
- Visits history tab - add ‘is’ to message displaying when no visits history
- Is it possible to remove final line from the tables in the tabs? E.g. in the screenshot of the VO tab - it’d look nicer if there wasn’t a final line under the table and just the tab outline
- Back button on page takes you back to search field but not search results


tpmcgowan (dismissed):
LGTM!

tpmcgowan (approved):

PR #203, opened 2022-05-26
Update visit session types and test data

tpmcgowan:
Update to reflect `restriction` property being removed from visit sessions.


AledWEvans (approved):
LGTM

PR #204, opened 2022-05-26
VB-810: Only show closed slots for visitors with closed restriction

tpmcgowan:
If a visitor with a CLOSED restriction is selected:
* set the `visitRestriction` in the session to be `CLOSED`
* set the reason for closed visits in the session to be `visitor`
* show an information banner on the visit date/time picker page to explain why only closed visits are being shown.

(Previously the `visitRestriction` was set to a default of `OPEN` at the start of the session; this is now set at the next stage in the journey when visitors are selected so there is a small refactor of the session validation code.)

### DIFF
--- a/server/middleware/sessionCheckMiddleware.ts
+++ b/server/middleware/sessionCheckMiddleware.ts
@@ -14,13 +14,15 @@ export default function sessionCheckMiddleware({ stage }: { stage: number }): Re
       !visitSessionData.prisoner.name ||
       !isValidPrisonerNumber(visitSessionData.prisoner.offenderNo || '') ||
       !visitSessionData.prisoner.dateOfBirth ||
-      !visitSessionData.prisoner.location ||
-      !visitSessionData.visitRestriction
+      !visitSessionData.prisoner.location
     ) {
       return res.redirect('/search/prisoner/?error=missing-prisoner')
     }
 
-    if (stage > 1 && (!visitSessionData.visitors || visitSessionData.visitors.length === 0)) {
+    if (
+      stage > 1 &&
+      (!visitSessionData.visitors || visitSessionData.visitors.length === 0 || !visitSessionData.visitRestriction)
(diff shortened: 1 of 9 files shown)


server/middleware/sessionCheckMiddleware.ts
```diff
@@ -14,13 +14,15 @@ export default function sessionCheckMiddleware({ stage }: { stage: number }): Re
...
-      !visitSessionData.visitRestriction
+      !visitSessionData.prisoner.location
     ) {
       return res.redirect('/search/prisoner/?error=missing-prisoner')
     }
 
-    if (stage > 1 && (!visitSessionData.visitors || visitSessionData.visitors.length === 0)) {
+    if (
+      stage > 1 &&
+      (!visitSessionData.visitors || visitSessionData.visitors.length === 0 || !visitSessionData.visitRestriction)
```
AledWEvans, line 24: Is this var correctly named visitSessionData, as it is not the same as a visistSession object from the server side?

philip-milne: For me it's not clear what these nested conditions are checking - maybe a function isValid.... would be cleaner.

tpmcgowan: @AledWEvans - the `visitSessionData` here is an object stored in the user's session that we use to build up options chosen in the UI during the journey (it's defined [here](https://github.com/ministryofjustice/book-a-prison-visit-staff-ui/blob/d551819121223705c19a3640334388dc90a276c3/server/%40types/bapv.d.ts#L174)). 

@philip-milne -  this middleware is basically just checking that the user has accumulated the correct bits and pieces in their session as they progress through stages of the journey. But, yes, you're right that this could be clearer, it's got a bit unwieldy as it's grown. Will have a chat to Paul about how we might refactor this as a separate task...

PR #205, opened 2022-05-27
VB-811: Option to choose open/closed visit type for prisoners with 'closed' restriction

tpmcgowan:
If the **prisoner** has a closed restriction then, after selecting visitors, show a page to choose visit type (open/closed) before proceeding to choose visit date/time. If a closed visit is chosen the reason for this ('prisoner') is stored to display in the subsequent banner.

If the **visitor** has a closed restriction then the visit type page is not used and the visit is automatically set to CLOSED.

### DIFF
--- a/server/routes/bookAVisit.ts
+++ b/server/routes/bookAVisit.ts
@@ -146,7 +146,10 @@ export default function routes(
       restriction => restriction.restrictionType === 'CLOSED'
     )
 
-    res.render('pages/visitType', { restrictions: closedRestrictions })
+    res.render('pages/visitType', {
+      restrictions: closedRestrictions,
+      visitors: visitSessionData.visitors,
+    })
   })
 
   get(
--- a/server/routes/bookAVisit.ts
+++ b/server/routes/bookAVisit.ts
@@ -139,6 +139,16 @@ export default function routes(
     }
   )
 
+  get('/visit-type', sessionCheckMiddleware({ stage: 2 }), async (req, res) => {
+    const { visitSessionData } = req.session
+
+    const closedRestrictions = visitSessionData.prisoner.restrictions.filter(
+      restriction => restriction.restrictionType === 'CLOSED'
+    )
+
+    res.render('pages/visitType', { restrictions: closedRestrictions })
+  })
+
   get(
     '/select-date-and-time',
     sessionCheckMiddleware({ stage: 2 }),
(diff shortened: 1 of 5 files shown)


server/routes/bookAVisit.ts
```diff
@@ -123,9 +125,50 @@ export default function routes(
         return closedVisit || visitor.restrictions.some(restriction => restriction.restrictionType === 'CLOSED')
       }, false)
       visitSessionData.visitRestriction = closedVisitVisitors ? 'CLOSED' : 'OPEN'
-
       visitSessionData.closedVisitReason = closedVisitVisitors ? 'visitor' : undefined
 
+      const closedVisitPrisoner = visitSessionData.prisoner.restrictions.reduce((closedVisit, restriction) => {
```
psoleckimoj, line 130: Might be clearer to use a `.some` rather than `.reduce`? Would make it clearer what it's doing?

server/routes/bookAVisit.ts
```diff
@@ -123,9 +125,50 @@ export default function routes(
...
+        return closedVisit || restriction.restrictionType === 'CLOSED'
+      }, false)
+
+      if (!closedVisitVisitors && closedVisitPrisoner) {
```
psoleckimoj, line 134: Could combine these into a ternary but that's up to you :) Means it's a single return statement.

psoleckimoj (dismissed):
LGTM, up to you if you want to make those changes but happy to merge as is.

server/routes/bookAVisit.ts
tpmcgowan: Good spot! Have updated to make it simpler.

server/routes/bookAVisit.ts
```diff
@@ -123,9 +125,48 @@ export default function routes(
         return closedVisit || visitor.restrictions.some(restriction => restriction.restrictionType === 'CLOSED')
       }, false)
       visitSessionData.visitRestriction = closedVisitVisitors ? 'CLOSED' : 'OPEN'
-
       visitSessionData.closedVisitReason = closedVisitVisitors ? 'visitor' : undefined
 
+      const closedVisitPrisoner = visitSessionData.prisoner.restrictions.some(restriction => {
+        return restriction.restrictionType === 'CLOSED'
+      })
```
psoleckimoj, line 132: Can actually do on 1 (2 after prettier) lines but I'm going to approve and merge anyways 🙂 

`const closedVisitPrisoner = visitSessionData.prisoner.restrictions.some(restriction => restriction.restrictionType === 'CLOSED')`

tpmcgowan: I should have spotted that as I've got something similar a couple of lines above 🤦‍♂️!

PR #212, opened 2022-06-09
VB-845, Safe function added

hutcheonb-moj:
## Description
Character's being escaped when main contact is displayed on page
Safe added to checkYourBooking.njk and confirmation.njk
Added to both version of the output

## Desired outcome
Names like O'brien Smith will no longer be escaped and display correctly


psoleckimoj (approved):
LGTM!

PR #214, opened 2022-06-10
VB-909: Check your answers page updates

psoleckimoj:
Not sure if this is the best way to do it. Had to use an array because visitors needs to run through a loop to build up visitors.


tpmcgowan (approved):
LGTM!

PR #216, opened 2022-06-14
Tidy up error page formatting to stop it looking broken

psoleckimoj:
Just makes it look less like the app is broken since the font, tags etc are all wrong.


tpmcgowan (approved):
LGTM!

PR #217, opened 2022-06-15
VB-933: Fix paging bug on visits search

psoleckimoj:


server/data/prisonerSearchClient.test.ts
```diff
@@ -102,5 +102,48 @@ describe('prisonSearchClientBuilder', () => {
         content: results,
       })
     })
+
+    it('should return data from api with the multiple pages', async () => {
+      const prisonerNumbers = []
+      const results = []
+      const numberOfResults = 15
```
tpmcgowan, line 109: Wonder if it would make sense here to do something like ` Math.ceil(config.apis.prisonerSearch.pageSize * 1.5)` rather than `15` as tests might fail if `pageSize` increased? Not changed via an env variable though so perhaps not a big issue...

tpmcgowan (approved):
LGTM!

server/data/prisonerSearchClient.test.ts
psoleckimoj: True, I'll put another PR in after to update 🙂

PR #224, opened 2022-06-22
Move to SQS

psoleckimoj:


server/services/auditService.test.ts
```diff
@@ -0,0 +1,81 @@
...
+  return {
+    SQSClient: jest.fn(() => sqsClientInstance),
+    SendMessageCommand,
+  }
+})
+
+describe('Audit service', () => {
+  let auditService: AuditService
+
+  describe('getVisitors', () => {
```
tpmcgowan, line 18: is `getVisitors` the right label for this?

server/routes/search.ts
```diff
@@ -53,6 +55,8 @@ export default function routes(
     const realNumberOfResults = errors.length > 0 ? 0 : numberOfResults
     const currentPageMax = parsedPage * pageSize
     const to = realNumberOfResults < currentPageMax ? realNumberOfResults : currentPageMax
+    auditService.prisonerSearch(search, 'HEI', res.locals.user?.username)
```
tpmcgowan, line 58: Is it worth adding to the test for this route a check that audit called?

tpmcgowan (dismissed):
LGTM - like how this is done! Just two minor comments.

server/services/auditService.test.ts
psoleckimoj: Whoops 😆

server/routes/search.ts
psoleckimoj: Will have a nosey!

PR #225, opened 2022-06-22
VB-910: Add beta banner to all ages

psoleckimoj:


server/views/partials/header.njk
```diff
@@ -37,3 +39,15 @@
...
   </span>
 </div>
+
+{% if not hidePhaseBanner %}
+  <div class="bapv-phase-banner-container">
+    {{ govukPhaseBanner({
+      tag: {
+        text: "beta"
+      },
+      html: 'This is a new service – your <a class="govuk-link" href="https://eu.surveymonkey.com/r/9ZBN3YJ">feedback</a> will help us to improve it.',
```
tpmcgowan, line 49: Does this need to open in a new tab - it's mentioned on the ticket (VB-496)...?

tpmcgowan (approved):
LGTM - just the query about new tab

server/views/partials/header.njk
psoleckimoj: I've left a comment on the ticket, it's not usually a good idea. https://www.w3.org/TR/WCAG20-TECHS/G200.html

PR #228, opened 2022-06-22
Prevent button double clicks

tpmcgowan:
Adding `preventDoubleClick: true` to existing button components (in advance of putting more buttons in for the dead-end journey points). Saw some double clicking going on while observing user research; figured we may as well add this attribute.


psoleckimoj (approved):
LGTM!

PR #232, opened 2022-06-28
VB-597: Add return to start button for journey dead-ends

tpmcgowan:
* Add a route (`/back-to-start`) that clears the session and redirects to `/`
* Add a Nunjucks component `backToStartButton.njk` that can be included to render a button with this link
* Add button to two journey 'dead-ends', where:
  * Prisoner has no approved visitors
  * No time/date slots available


psoleckimoj (approved):
LGTM!

PR #236, opened 2022-07-04
Date/time picker: always show day and time filter and tweak text

tpmcgowan:
Previously, day/time filter wasn't shown if no slots available - creating a 'dead-end'. This change makes it always display.


hutcheonb-moj (approved):
LGTM

PR #237, opened 2022-07-05
VB-597: Select visitors: handle only banned/child dead-end

tpmcgowan:
This fixes a journey 'dead-end' on the Select Visitors page where a booking can't proceed if only banned or child visitors are listed. Instead of showing the Continue button, the back-to-start button is displayed.


hutcheonb-moj (approved):
LGTM

PR #240, opened 2022-07-13
VB-417: User audit - fixes, tweaks and tests

tpmcgowan:
* adds a middleware to get the AppInsights operation ID available in each request to be logged
* minor fixes to what gets logged to audit
* expand test coverage


hutcheonb-moj (approved):
LGTM

PR #243, opened 2022-07-15
VB-1026: check booking route - bug fix and tests

tpmcgowan:
* fixes an issue where a confirmation SMS would still have been sent if the visit booking **failed**
* add test coverage for this route


hutcheonb-moj (approved):
LGTM

PR #244, opened 2022-07-15
VB-1013, still WIP - Client+Service ++ Tests done

hutcheonb-moj:
sendSms changed to sendBookedSms so cancellation sms can be built
NotificationsApiClient - Cancellation notification added, including tests
NotificationService - Cancellation notification added, including tests

Cancellation template added to config.ts

Routes - Book A Visit had to be updated to reflect the changing of name of sendSms


tpmcgowan (approved):
LGTM!

PR #245, opened 2022-07-18
VB-417:  Add more user audit events

tpmcgowan:
* Added new audit events:
  * `SEARCHED_VISITS`
  * `VIEWED_VISIT_DETAILS`
  * `VISIT_RESTRICTION_SELECTED`
* Removed `PAGE_VIEW` as not used and no longer required
* Documented audit events and how to use Application Insights to investigate them
* Minor tidy / bug fixes

### DIFF
--- a/server/routes/bookAVisit.ts
+++ b/server/routes/bookAVisit.ts
@@ -173,6 +173,14 @@ export default function routes(
       visitSessionData.visitRestriction = req.body.visitType
       visitSessionData.closedVisitReason = req.body.visitType === 'CLOSED' ? 'prisoner' : undefined
 
+      await auditService.visitRestrictionSelected(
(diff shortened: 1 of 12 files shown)


server/routes/bookAVisit.ts
```diff
@@ -262,7 +270,7 @@ export default function routes(
         visitSessionData.visitStatus = visitStatus
       }
 
-      auditService.reservedVisit(
+      await auditService.reservedVisit(
```
tpmcgowan, line 273: I've added `await` wherever `auditService` methods are called as it it looked like they should be there to be consistent with how our other services are called?

psoleckimoj (approved):
LGTM!

PR #254, opened 2022-07-22
VB-952: Handle missing prisoner data (e.g. prisoner transferred or released)

tpmcgowan:
Viewing a visit or searching by visit reference could fail if the prisoner had transferred from Hewell or been released as the API endpoints used to look up prisoner details (e.g. name) were tied down to Hewell (as a requirement of the booking journey).

This change adds the `GET /prisoner/{id}` endpoint from the `global-search-resource` in Prisoner Offender Search API and uses this to retrieve prisoner details - but **only** when viewing a visit or searching by visit reference in order to keep booking journey behaviour unchanged.

(Until further work is done on requirements regarding prisoner transfer/release, the prisoner location on the visit summary page will show **Unknown** if the prisoner is no longer at Hewell.)

### DIFF
--- a/server/routes/visit.test.ts
+++ b/server/routes/visit.test.ts
@@ -135,7 +135,7 @@ describe('GET /visit/:reference', () => {
 
     const additionalSupport = ['Wheelchair ramp', 'custom request']
 
-    prisonerSearchService.getPrisoner.mockResolvedValue(prisoner)
+    prisonerSearchService.getPrisonerById.mockResolvedValue(prisoner)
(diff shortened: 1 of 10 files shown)


server/routes/visit.test.ts
```diff
@@ -332,6 +248,37 @@ describe('GET /visit/:reference', () => {
       })
   })
 
+  // Temporarily hiding any locations other than Hewell pending more work on transfer/release (see VB-907, VB-952)
```
tpmcgowan, line 251: This new test is actually the only real change to this file; pretty much all the above is just me refactoring slightly to remove duplicate test data.

PR #259, opened 2022-07-26
VB-1033: Don't sent audit event if empty search

tpmcgowan:
Empty search requests were no longer calling the search API but **were** still logging an audit event. This change only logs audit events for valid search terms (i.e. currently 2 characters or more).


psoleckimoj (approved):
LGTM!

PR #261, opened 2022-07-27
VB-927: Fix issue where closed visits don't display on visits page if…

psoleckimoj:
… there are no open visits

### DIFF
--- a/server/services/visitSessionsService.ts
+++ b/server/services/visitSessionsService.ts
@@ -16,9 +16,10 @@ import VisitSchedulerApiClient from '../data/visitSchedulerApiClient'
 import WhereaboutsApiClient from '../data/whereaboutsApiClient'
 import { VisitSession, Visit, SupportType, OutcomeDto } from '../data/visitSchedulerApiTypes'
 import { ScheduledEvent } from '../data/whereaboutsApiTypes'
-import { prisonerDateTimePretty, prisonerTimePretty, sortByTimestamp } from '../utils/utils'
+import { prisonerDateTimePretty, prisonerTimePretty } from '../utils/utils'
 import PrisonerContactRegistryApiClient from '../data/prisonerContactRegistryApiClient'
 import buildVisitorListItem from '../utils/visitorUtils'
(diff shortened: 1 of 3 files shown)


server/services/visitSessionsService.ts
```diff
@@ -389,7 +329,7 @@ export default class VisitSessionsService {
       startTimestamp: visit.startTimestamp,
       visitDate: prisonerDateTimePretty(visit.startTimestamp),
       visitTime,
-      visitRestriction: visit.visitRestriction,
+      visitRestriction: visit.visitRestriction === 'OPEN' ? 'OPEN' : 'CLOSED',
```
psoleckimoj, line 332: This is because imported visits are sometimes UNKNOWN, not sure what we should default to tbh.

psoleckimoj: Have asked the question, this may change to OPEN

server/utils/visitsUtils.ts
```diff
@@ -0,0 +1,67 @@
...
+): {
+  openSlots: VisitsPageSlot[]
+  closedSlots: VisitsPageSlot[]
+  firstSlotTime: string
+} => {
+  const openSlots: VisitsPageSlot[] = []
+  const closedSlots: VisitsPageSlot[] = []
+
+  visits.forEach((visit: ExtendedVisitInformation) => {
+    if (visit.visitRestriction === 'OPEN') {
```
psoleckimoj, line 15: Again this may change to OPEN or UNKNOWN

server/services/visitSessionsService.ts
tpmcgowan: Similar to what's on https://dsdmoj.atlassian.net/browse/VB-995 - I raised the question of what to do with 'Unknown' visits. Feels like that issue should be linked to VB-927?

psoleckimoj: @tpmcgowan yep, Phoebe is going to link them as it all relates to how to display them.

tpmcgowan (approved):
LGTM

PR #264, opened 2022-07-28
VB-543: Incorporate non-associations into date picker page

tpmcgowan:
Slots that shouldn't be available because of non-associations can be filtered out of the available list by passing `prisonerId` to the visit scheduler when requesting the list of available slots.


psoleckimoj (approved):
LGTM!

PR #265, opened 2022-07-29
Vb 597 back to manage bugfix

hutcheonb-moj:
Work in progress

### DIFF
--- a/server/routes/bookAVisit.ts
+++ b/server/routes/bookAVisit.ts
@@ -216,7 +216,6 @@ export default function routes(
         }
         return true
       })
-      logger.info(slotsPresent)
 
       req.session.slotsList = slotsList
       req.session.timeOfDay = timeOfDay
--- a/server/routes/bookAVisit.ts
+++ b/server/routes/bookAVisit.ts
@@ -209,13 +209,6 @@ export default function routes(
       }
       let slotsPresent = false
 
-      Object.entries(slotsList).find(([, value]) => {
-        if (value.length !== 0) {
-          slotsPresent = true
-          return false
-        }
-        return true
-      })
       slotsPresent = Object.values(slotsList).some(value => value.length)
 
       req.session.slotsList = slotsList
--- a/server/routes/bookAVisit.ts
+++ b/server/routes/bookAVisit.ts
@@ -216,6 +216,7 @@ export default function routes(
         }
         return true
       })
+      slotsPresent = Object.values(slotsList).some(value => value.length)
 
       req.session.slotsList = slotsList
       req.session.timeOfDay = timeOfDay
--- a/server/routes/bookAVisit.ts
+++ b/server/routes/bookAVisit.ts
@@ -207,9 +207,8 @@ export default function routes(
       if (!Object.keys(formValues).length && visitSessionData.visit?.id) {
         formValues['visit-date-and-time'] = visitSessionData.visit?.id
       }
-      let slotsPresent = false
 
-      slotsPresent = Object.values(slotsList).some(value => value.length)
+      const slotsPresent = Object.values(slotsList).some(value => value.length)
 
       req.session.slotsList = slotsList
       req.session.timeOfDay = timeOfDay
--- a/server/routes/bookAVisit.ts
+++ b/server/routes/bookAVisit.ts
@@ -207,6 +207,16 @@ export default function routes(
       if (!Object.keys(formValues).length && visitSessionData.visit?.id) {
         formValues['visit-date-and-time'] = visitSessionData.visit?.id
       }
+      let slotsPresent = false
+
+      Object.entries(slotsList).find(([, value]) => {
+        if (value.length !== 0) {
+          slotsPresent = true
+          return false
+        }
+        return true
+      })
+      logger.info(slotsPresent)
 
       req.session.slotsList = slotsList
       req.session.timeOfDay = timeOfDay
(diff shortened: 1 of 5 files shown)


server/views/pages/dateAndTime.njk
```diff
@@ -151,6 +151,15 @@
     </form>
 
     {% if slotsList | length %}
+      {% for i in slotsList | list %}
```
psoleckimoj, line 154: Can you do this processing in the route and pass in `slotsPresent`? Keeps as much processing out of the templates as possible. Ideally the template just wants to display stuff rather than working out variables. Also use meaningful variable names, `i` could be `slot` as a more descriptive name 👍

psoleckimoj: Could then probably use some shiny array methods like `.find()` etc to do what you want more functionally 🙂

server/services/visitSessionsService.test.ts
```diff
@@ -530,19 +530,7 @@ describe('Visit sessions service', () => {
...
-              morning: [],
-              afternoon: [],
-            },
-            slots: {
-              morning: [],
-              afternoon: [],
-            },
-          },
-        ],
+        'February 2022': [],
```
tpmcgowan, line 533: I'm thinking that if you can get it to just return `{}` for this case then in the template the existing `{% if slotsList | length %}` check would work and so no changes would need doing in the template itself 🧐?

tpmcgowan: My logic being that we're not really doing what the test case says (i.e. "no results") for either the original or what you've changed it to. 

If trying this, probably need to check it does what's expected with multiple months: if, say, February has only a morning slot and March only an afternoon one that should return nothing for February but some data for March. Might need new test or tweak to an existing one.

tpmcgowan: For info, there's also [dateAndTime.test.ts](https://github.com/ministryofjustice/book-a-prison-visit-staff-ui/blob/main/server/views/pages/dateAndTime.test.ts) where there are some tests for this template in isolation that might be handy.

hutcheonb-moj: This whole issue arises because its difficult to reduce the data to {} in this case, do you think it's really important to try and get it to that stage?

server/routes/bookAVisit.ts
```diff
@@ -207,6 +207,16 @@ export default function routes(
...
+      let slotsPresent = false
+
+      Object.entries(slotsList).find(([, value]) => {
+        if (value.length !== 0) {
+          slotsPresent = true
+          return false
+        }
+        return true
+      })
+      slotsPresent = Object.values(slotsList).some(value => value.length)
```
tpmcgowan, line 219: do you need the section above now you've added this line?

hutcheonb-moj: well spotted I had it folded in my view...

tpmcgowan: 👍 can probably just do `const slotsPresent = Object....` rather than `let` as you don't re-assign it.

server/services/visitSessionsService.test.ts
tpmcgowan: No, not necessarily - was just thinking of an alternative approach to this which would have been tidying the object returned from the service to remove empty months. E.g. in `visitSessionService.ts` around line 161 where you have:
```
if (visitSlotsforDay.slots.morning.length > 0 || visitSlotsforDay.slots.afternoon.length > 0) {
...
```
and an `else` block, something like:
```
        if (visitSlotsforDay.slots.morning.length > 0 || visitSlotsforDay.slots.afternoon.length > 0) {
          allVisitSlots.push(visitSlotsforDay)
          availableSessions[month] = allVisitSlots
        } else {
          delete availableSessions[month]
        }
      })
    })
```
Untested - and not saying this is particularly a better idea! This data's a bit fiddly however you do it...

server/routes/bookAVisit.ts
hutcheonb-moj: got this implemented! much cleaner data

server/services/visitSessionsService.ts
```diff
@@ -158,11 +158,13 @@ export default class VisitSessionsService {
           const timeEnd = parse(`${visitSlotsforDay.date} ${year} 23:59`, 'EEEE d MMMM yyyy H:mm', new Date())
           visitSlotsforDay.prisonerEvents.afternoon = getPrisonerEvents(prisonerEvents, timeStart, timeEnd)
         }
-
-        allVisitSlots.push(visitSlotsforDay)
+        if (visitSlotsforDay.slots.morning.length > 0 || visitSlotsforDay.slots.afternoon.length > 0) {
```
psoleckimoj, line 161: I'm not sure this is in the right place, it's essentially saying if there are no sessions for this day, remove the month. You can see the `availableSessions[month] = allVisitSlots` has moved inside the inner loop. The `if` and the following line are correct however the other 2 lines need to happen after the `availableSessions[month].forEach` loop with a different condition on the `if`. So the `if` would maybe be `if (availableSessions[month].length`?

The `if` for the `visitSlotsforDay` is correct, but the one for `availableSessions[month]` should be after the `availableSessions[month]` loop.

hutcheonb-moj: re-implemented this outside of the month's loop in my most recent commit

psoleckimoj (approved):
LGTM!

PR #266, opened 2022-07-29
VB-1034: disallow double booking

tpmcgowan:
Incorporates the `sessionConflicts` information from the visit scheduler to disable slots flagged as a `DOUBLE_BOOKING`, meaning that the prisoner already has a visit at this time.

The `DOUBLE_BOOKING` flag is also set once a visit is `RESERVED` so it checks whether the slot flagged is the one being managed by the current journey, in which case it is not disabled.


psoleckimoj (approved):
LGTM!

PR #276, opened 2022-08-12
VB-1079: Refactor select date and time for amend/book

psoleckimoj:


server/views/pages/visit/dateAndTime.njk
```diff
@@ -0,0 +1,262 @@
+{% extends "layout.njk" %}
```
tpmcgowan, line 1: We've got a `dateAndTime.test.ts` that goes with the corresponding `bookAVisit` version of this template file - I think originally done to test the template on its own as there's so much going on in it. Not sure whether we still need it and/or if it's worth duplicating it here too.... 🤔

tpmcgowan (approved):
LGTM!

server/views/pages/visit/dateAndTime.njk
psoleckimoj: I think it will be, but as you say might be best to refactor beforehand or as part of writing the tests for this. Will look at in the next phase 🙂

PR #279, opened 2022-08-17
Remove duplicate call to contact registry on prisoner profile

tpmcgowan:
Prisoner profile page was calling contact registry both during getting past visits and upcoming visits. This charge:
* adds test data to cover contact registry and visit data
* refactors to only call the contact registry once for the profile page


psoleckimoj (approved):
LGTM!

PR #280, opened 2022-08-18
VB-1079: Amend date and time

psoleckimoj:


server/@types/bapv.d.ts
```diff
@@ -140,6 +140,7 @@ export type VisitSlot = {
   endTimestamp: string
   availableTables: number
   visitRoomName: string
+  visitRestriction: string
```
psoleckimoj, line 143: Required to know what sort of slot this is for amend journey as this can't be assumed from the availableTables which is based off whichever is the current restriction.

server/@types/bapv.d.ts
```diff
@@ -179,9 +180,11 @@ export type VisitSessionData = {
     dateOfBirth: string
     location: string
     restrictions?: OffenderRestriction[]
+    previousRestrictions?: OffenderRestriction[]
   }
   visit?: VisitSlot
   visitRestriction?: Visit['visitRestriction']
+  previousVisitRestriction?: Visit['visitRestriction']
```
psoleckimoj, line 187: Used to work out whether a visit restriction has changed in an amend journey.

server/data/visitSchedulerApiClient.ts
```diff
@@ -137,6 +137,15 @@ class VisitSchedulerApiClient {
...
 
+  startAmendVisit(visitReference: string): Promise<Visit> {
+    return this.restclient.put({
+      path: `/visits/${visitReference}`,
+      data: <UpdateVisitRequestDto>{
+        visitStatus: 'RESERVED',
+      },
+    })
+  }
+
```
psoleckimoj, line 148: Only required to update the status when doing this, current update function sends everything, could maybe changed to update status only?

server/routes/visit.ts
```diff
@@ -84,11 +84,12 @@ export default function routes(
         location: prisoner.locationDescription,
       },
       visit: {
-        id: visit.reference,
+        id: '',
```
psoleckimoj, line 87: id is not reference, this id is generated by the code and is always from 1 upwards depending on available sessions.

server/routes/visit.ts
```diff
@@ -84,11 +84,12 @@ export default function routes(
...
         startTimestamp: visit.startTimestamp,
         endTimestamp: visit.endTimestamp,
         availableTables: 0,
         visitRoomName: visit.visitRoom,
+        visitRestriction: visit.visitRestriction,
```
psoleckimoj, line 92: Restriction as time of original booking

server/routes/visit.ts
```diff
@@ -104,6 +108,8 @@ export default function routes(
 
     req.session.amendVisitSessionData = Object.assign(req.session.amendVisitSessionData ?? {}, visitSessionData)
 
+    // await visitSessionsService.startAmendVisit({ username: res.locals.user?.username, visitReference: reference })
+
```
psoleckimoj, line 112: Commented out while we wait to hear what we need to do about reserving visits.

server/data/visitSchedulerApiClient.ts
tpmcgowan: Just wondering about the `RESERVED` here - as I think you've mentioned if amend is started and then abandoned for whatever reason the visit might automatically housekept! 

If we're re-using the same visit, it feels like there should perhaps be another status, e.g. `AMENDING` or something - this could signify what's going on; stop it simultaneously being amended by someone else; and, cron job could revert any it finds to `BOOKED` were they to be abandoned 🤔 ? Obviously would need some much wider discussion!

psoleckimoj: @tpmcgowan yep, the code that calls this is currently commented out for the reason you say 🙂

server/routes/visitorUtils.ts
```diff
@@ -11,6 +11,19 @@ export const getSelectedSlot = (slotsList: VisitSlotList, selectedSlot: string):
     .find(slot => slot.id === selectedSlot)
 }
 
+export const getSelectedSlotByStartTimestamp = (
```
tpmcgowan, line 14: If only these slots came with their own ID numbers....!

psoleckimoj: 😆

tpmcgowan (approved):
LGTM!

PR #284, opened 2022-08-19
Add initial tests pulled from bookAVisit journey

psoleckimoj:


server/routes/visitJourney/visitType.test.ts
```diff
@@ -0,0 +1,184 @@
+import type { Express } from 'express'
```
psoleckimoj, line 1: Added with a skip, but would like the others merging before I try and separate this out. Can remove this from the commit if necessary.

tpmcgowan (approved):
LGTM

PR #289, opened 2022-08-25
vb-1112, remove startUpdateVisit

hutcheonb-moj:
## Purpose
Remove startUpdateVisit and all of its references

## Reason
The process around update visit has changed, we will no longer be changing the previous booking to reserved as this would mean the entire booking would be lost (to housekeeping) should this visit not be re-booked


psoleckimoj (approved):
lgtm!

PR #296, opened 2022-08-26
VB-1112 Reserve new slot on update

tpmcgowan:
This change ensures that for an update visit journey a new visit is created each time, with the existing visit untouched. (Future work will 'swap' these over by booking the new one and cancelling the previous one.)

Most changes relate to refactoring to simplify by removing separate storing of old/new visit details in session.

To do in subsequent changes for this ticket:
* sort issues with table count showing negative values in some update scenarios
* tests for new visit creation

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
   "lint-staged": {
     "*.{ts,js,css}": [
       "prettier --write",
-      "eslint --fix"
+      "eslint --fix --max-warnings 0"
     ],
     "*.json": [
       "prettier --write"
--- a/server/@types/bapv.d.ts
+++ b/server/@types/bapv.d.ts
@@ -194,6 +194,7 @@ export type VisitSessionData = {
     contactName?: string
   }
   visitReference?: string
+  previousVisitReference?: string
   visitStatus?: Visit['visitStatus']
 }
 
--- a/server/services/visitSessionsService.ts
+++ b/server/services/visitSessionsService.ts
@@ -174,11 +174,11 @@ export default class VisitSessionsService {
   async updateVisit({
     username,
     visitData,
-    visitStatus = 'RESERVED',
+    visitStatus,
   }: {
     username: string
     visitData: VisitSessionData
-    visitStatus?: string
+    visitStatus: Visit['visitStatus']
   }): Promise<Visit> {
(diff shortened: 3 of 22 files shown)


package.json
```diff
@@ -79,7 +79,7 @@
   "lint-staged": {
     "*.{ts,js,css}": [
       "prettier --write",
-      "eslint --fix"
+      "eslint --fix --max-warnings 0"
```
tpmcgowan, line 82: Changing this so that settings for pre-commit hook match settings run during build (I got caught out by it passing pre-commit hooks but then failing on CircleCI...)

server/views/pages/bookAVisit/dateAndTime.njk
```diff
@@ -21,7 +21,7 @@
     {%- set tableText %}
       {% if (doubleBooked and not checked) -%}
         Prisoner has a visit
-      {%- elseif slot.availableTables === 0 -%}
+      {%- elseif slot.availableTables <= 0 -%}
```
tpmcgowan, line 24: Handle availability potentially being temporarily below zero while a visit is being updated.

server/services/visitSessionsService.ts
```diff
@@ -174,11 +174,11 @@ export default class VisitSessionsService {
   async updateVisit({
     username,
     visitData,
-    visitStatus = 'RESERVED',
+    visitStatus,
```
tpmcgowan, line 177: Stop `updateVisit` defaulting to `RESERVED` to avoid inadvertently 'downgrading' a booking - that could then get housekept.

server/@types/bapv.d.ts
```diff
@@ -194,6 +194,7 @@ export type VisitSessionData = {
     contactName?: string
   }
   visitReference?: string
+  previousVisitReference?: string
```
psoleckimoj, line 197: We might be able to remove the other `previous*` fields if we load the existing data into the session?

tpmcgowan: Yeah, I think it _could_ work without them - but I think they are needed because of the part where it says restrictions have changed so open/closed may need to alter.

psoleckimoj: I flip-flopped and think we need it.

PR #297, opened 2022-08-30
VB-1126 Tech Debit : remove VisitStatus from call to create visit in …

AledWEvans:
Remove VisitStatus from call to create visit.

In order to remove api Elements that is not part of POC.


tpmcgowan (approved):
LGTM!

PR #298, opened 2022-08-30
VB-1120: On update go directly to select visitors page

tpmcgowan:
Remove the 'What do you want to amend' page so that clicking 'Update' a visit goes straight to the first page in the journey, which is the select visitors page.


hutcheonb-moj (approved):
LGTM

PR #299, opened 2022-08-31
VB-1072: Clear session after visit cancellation

tpmcgowan:
Now that viewing booking summary populates session, ensure this is cleared once a visit is cancelled.


psoleckimoj (approved):
LGTM!

PR #301, opened 2022-09-01
VB-1112: de-duplicate templates

tpmcgowan:
Consolidate the visit journey templates so there's only one of each for book and update.


server/routes/visitJourney/visitJourneyUtils.ts
```diff
@@ -0,0 +1,6 @@
+// return URL prefix for either visit booking or visit update journey
+const getUrlPrefix = (isUpdate: boolean, previousVisitReference?: string) => {
```
tpmcgowan, line 2: Was in two minds whether or not to create yet another 'utils' file but figured there's likely to be other things here to support these 'visitJourney' classes - and these various utils files probably need sorting out at some point anyway...

server/routes/visitJourney/visitJourneyUtils.ts
```diff
@@ -0,0 +1,6 @@
+// return URL prefix for either visit booking or visit update journey
+const getUrlPrefix = (isUpdate: boolean, previousVisitReference: string) => {
```
tpmcgowan, line 2: Was in two minds whether or not to create yet another 'utils' file but figured there's likely to be other things here to support these 'visitJourney' classes - and these various utils files probably need sorting out at some point anyway...

server/routes/visitJourney/confirmation.ts
```diff
@@ -5,7 +5,7 @@ export default class Confirmation {
   constructor(private readonly mode: string) {}
 
   async get(req: Request, res: Response): Promise<void> {
-    const isUpdate = this.mode === 'update'
+    // const isUpdate = this.mode === 'update'
```
tpmcgowan, line 8: Left commented out to avoid lint errors, as this will be needed when confirmation page takes account of booking vs update

psoleckimoj (approved):
LGTM!

PR #304, opened 2022-09-02
VB-1137, use url prefix in back button

hutcheonb-moj:
## Description

Back button was defaulting to the visit journey URL's, this PR aim's to use urlPrefix in place of a hardcoded url, so for both journeys the back button navigates the user to the correct page.

urlPrefix is already being used for the form actions, so this is a good re-use of existing code we've implemented.

### DIFF
--- a/server/routes/visitJourney/selectVisitors.ts
+++ b/server/routes/visitJourney/selectVisitors.ts
@@ -37,6 +37,13 @@ export default class SelectVisitors {
       formValues.visitors = visitSessionData.visitors.map(visitor => visitor.personId.toString())
     }
 
+    let returnAddress
+    if (isUpdate) {
+      returnAddress = `/visit/${visitSessionData.previousVisitReference}`
+    } else {
+      returnAddress = `/prisoner/${offenderNo}`
+    }
+
     res.render('pages/bookAVisit/visitors', {
       errors: req.flash('errors'),
       offenderNo: visitSessionData.prisoner.offenderNo,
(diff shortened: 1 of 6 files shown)


tpmcgowan:
Looks good so far 👍 . Might need to look at what happens with select visitors page, `server/views/pages/bookAVisit/visitors.njk`, as first time round this should go to `/prisoner/{offenderNo}` but if you come on an amend journey I guess it should go back the booking summary page 🤔

psoleckimoj (approved):
LGTM for these 4 pages. Can do the start and confirmation pages as separate PRs

hutcheonb-moj:
I finished the start page before I saw your comment on separate PR @psoleckimoj 
However the confirmation page doesn't include a back button as it has the option to return to any section of the journey to make a change, so I think this ticket is now complete

server/routes/visitJourney/selectVisitors.ts
```diff
@@ -37,6 +37,13 @@ export default class SelectVisitors {
       formValues.visitors = visitSessionData.visitors.map(visitor => visitor.personId.toString())
     }
 
+    let returnAddress
+    if (isUpdate) {
+      returnAddress = `/visit/${visitSessionData.previousVisitReference}`
+    } else {
+      returnAddress = `/prisoner/${offenderNo}`
+    }
+
```
psoleckimoj, line 46: A nicer way to write this would be:

```
const returnAddress = isUpdate ? `/visit/${visitSessionData.previousVisitReference}` : `/prisoner/${offenderNo}`
```

psoleckimoj: Otherwise good 👍

psoleckimoj (approved):
LGTM!

PR #305, opened 2022-09-07
VB-1162: Fix unknown visit type

psoleckimoj:


docker-compose.yml
```diff
@@ -1,11 +1,11 @@
 version: '3.1'
 services:
 
-  redis-bapv:
+  redis:
```
psoleckimoj, line 4: Accidental change but makes more sense when you're trying to run other projects that use redis at the same time and use `redis` as the service name.

hutcheonb-moj (approved):
LGTM

PR #307, opened 2022-09-07
vb-1080, additional support wasn't being updated if changed to none

hutcheonb-moj:
## Description
Attempting to fix the bug where additional support stays populated on both update journey and pre-booking changes.
When an additional support item has previously been selected, changing to none doesn't update the additional support in visitSessionData

### DIFF
--- a/server/routes/bookAVisit.test.ts
+++ b/server/routes/bookAVisit.test.ts
@@ -1799,7 +1799,8 @@ describe('POST /book-a-visit/additional-support', () => {
       })
   })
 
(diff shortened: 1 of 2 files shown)


psoleckimoj (approved):
LGTM

Could add a test to make sure this is cleared when no is selected?

psoleckimoj (dismissed):
LGTM!

psoleckimoj (approved):
LGTM

tpmcgowan (approved):
LGTM!

PR #308, opened 2022-09-08
vb-1169, incorporated booked endpoint into search by reference

hutcheonb-moj:
## Description
Added getBookedVisits endpoint, only shows booked visits instead of all visits.
Changed search to now use getBookedVisits endpoint, as being able to search for a visit and then change that visit after it has been cancelled or superseded causes errors.

Linked to VB-1168

### DIFF
--- a/server/services/visitSessionsService.ts
+++ b/server/services/visitSessionsService.ts
@@ -216,6 +216,16 @@ export default class VisitSessionsService {
     return this.buildVisitInformation(visit)
   }
 
+  async getBookedVisit({ username, reference }: { username: string; reference: string }): Promise<VisitInformation> {
+    const token = await this.systemToken(username)
+    const visitSchedulerApiClient = this.visitSchedulerApiClientBuilder(token)
+
+    logger.info(`Get visit ${reference}`)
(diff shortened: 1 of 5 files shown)


server/services/visitSessionsService.ts
```diff
@@ -216,6 +216,16 @@ export default class VisitSessionsService {
     return this.buildVisitInformation(visit)
   }
 
+  async getBookedVisit({ username, reference }: { username: string; reference: string }): Promise<VisitInformation> {
+    const token = await this.systemToken(username)
+    const visitSchedulerApiClient = this.visitSchedulerApiClientBuilder(token)
+
+    logger.info(`Get visit ${reference}`)
```
psoleckimoj, line 223: Only change I can see is the log message 👍

psoleckimoj (approved):
LGTM!

server/services/visitSessionsService.ts
hutcheonb-moj: Annoyingly I did consider that, then wasn't sure if it made sense from a log perspective!

PR #310, opened 2022-09-09
vb-1172-disable-update-and-cancel-on-unavailable

hutcheonb-moj:
## Description
Disable the buttons on the visit summary page when a visit has been superseded or cancelled already.
This work may be replaced when a design has been made but this ticket works for MVP.

On Superseded it displays the visit reference number for the new visit also links to the new visit and gives the new visit reference

### DIFF
--- a/server/views/pages/visit/summary.njk
+++ b/server/views/pages/visit/summary.njk
@@ -85,6 +85,7 @@
           }) }}
         </div>
       {% else %}
+      
         {% if visit.outcomeStatus === 'SUPERSEDED_CANCELLATION' %}
           <h2 class="govuk-heading-m">This visit has been {{ visit.visitNotes[0].text}} </h2>
         {% else %}
--- a/server/views/pages/visit/summary.njk
+++ b/server/views/pages/visit/summary.njk
@@ -62,27 +62,36 @@
           <span data-test="visit-phone">{{ visit.visitContact.telephone | formatTelephone }}</span>
         </li>
       </ul>
-
-      <div class="govuk-button-group">
-        {% set cancelButtonClasses = 'govuk-!-margin-top-5' %}
-        {% if updateJourneyEnabled %}
-          {% set cancelButtonClasses = 'govuk-!-margin-top-5 govuk-button--secondary' %}
+      
+      {% if visit.visitStatus === "BOOKED" %}
+        <div class="govuk-button-group">
+          {% set cancelButtonClasses = 'govuk-!-margin-top-5' %}
+          {% if updateJourneyEnabled %}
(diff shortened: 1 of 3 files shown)


tpmcgowan (changes requested):
Marking this as requiring changes as Richard's confirmed that the ticket should have been written to disable rather than remove the buttons. 

As discussed, also need to consider displaying the outcome status 'free text'  reason as well as just the status.

server/views/pages/visit/summary.njk
```diff
@@ -67,22 +68,63 @@
         {% set cancelButtonClasses = 'govuk-!-margin-top-5' %}
         {% if updateJourneyEnabled %}
           {% set cancelButtonClasses = 'govuk-!-margin-top-5 govuk-button--secondary' %}
+
```
hutcheonb-moj, line 71: Many if else blocks due to updateJourney feature flag being used. This can be made to be more concise and clear once we no longer require this flag

server/views/pages/visit/summary.njk
```diff
@@ -67,22 +68,63 @@
...
+              attributes: { "data-test": "update-visit" },
+              preventDoubleClick: true
+            }) }}
+          {% else %}
+            {{ govukButton({
+              text: "Update booking",
+              classes: "govuk-!-margin-top-5",
+              attributes: { "data-test": "update-visit" },
+              preventDoubleClick: true,
+              disabled: true
```
psoleckimoj, line 86: You can just put the condition on a single button. if it's disabled then you can't use it anyway so all you need to do it set `disabled` based on `visit.visitStatus === 'BOOKED'` so could do `disabled: visit.visitStatus === 'BOOKED'`?

server/views/pages/visit/summary.njk
```diff
@@ -67,22 +68,63 @@
...
+            }) }}
+          {% endif %}
+        {% endif %}
+
+        {% if visit.visitStatus === 'BOOKED' %}
```
psoleckimoj, line 91: Again I believe you can do the same here unless it's doing something odd when the button has a `href` and isn't disabling it?

server/views/pages/visit/summary.njk
tpmcgowan: Yeah - we tried that originally and it _seemed_ to work but it only visually disabled them (i.e. greyed out with CSS). You could still click on them! Found out because we couldn't figure out why the tests we'd written weren't working 🤦‍♂️!

psoleckimoj (approved):
Ignore my comments about the buttons, fault GDS code!
LGTM!

tpmcgowan (approved):

PR #311, opened 2022-09-10
Just use visit ref for SUPERSEDED_CANCELLATION reason text

tpmcgowan:
Change the recorded outcome status to be simply the visit reference so it can be more easily processed, e.g. to be re-used to form a URL. Message to the user audit log is unchanged.


hutcheonb-moj (approved):
LGTM, thanks Tom! Need this for my work on VB-1172

PR #315, opened 2022-09-12
Add test coverage, include in build artifacts and document

tpmcgowan:
Run test coverage report by default (only seems to add a couple of seconds). Include the reports in build artefacts on CIrcleCI for reference. Document where to find these.


psoleckimoj (approved):
LGTM!

PR #317, opened 2022-09-12
VB-1084: Standardise back to start button text

tpmcgowan:
Use same wording for buttons in various places that take user back to the 'home' page. (Arose as part of discussions on VB-1084.)


psoleckimoj (approved):
LGTM!

PR #325, opened 2022-09-14
vb-1196, autocomplete added to search for visit, works correctly

hutcheonb-moj:
## Description

Added autocomplete off to the search by visit reference field.

Autocomplete off on Edge doesn't have any effect on the main contact page (which this ticket was created for) - so will discuss this in stand up to see if we can agree on a course of action


tpmcgowan (approved):
LGTM!

PR #327, opened 2022-09-21
vb-1181, Remove visited state from all links

hutcheonb-moj:
## Description
Changed all visited links to default blue, added to top of local.scss and targeted a parent class to make sure it was targeting only the links, not the buttons

### DIFF
--- a/assets/sass/local.scss
+++ b/assets/sass/local.scss
@@ -1,3 +1,7 @@
+.govuk-main-wrapper a:not(.govuk-button):visited {
+  color: $govuk-link-colour;
+}
+
 .govuk-main-wrapper {
   min-height: 600px;
 }
--- a/assets/sass/local.scss
+++ b/assets/sass/local.scss
@@ -1,3 +1,7 @@
+a:visited {
+  color: $govuk-link-colour!important;
+}
+
 .govuk-main-wrapper {
   min-height: 600px;
 }
--- a/assets/sass/local.scss
+++ b/assets/sass/local.scss
@@ -1,7 +1,3 @@
-a:visited {
-  color: $govuk-link-colour!important;
-}
-
 .govuk-main-wrapper {
   min-height: 600px;
 }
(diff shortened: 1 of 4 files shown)


assets/sass/local.scss
```diff
@@ -1,3 +1,7 @@
+a:visited {
+  color: $govuk-link-colour!important;
```
psoleckimoj, line 2: Don't use `important`, add a new class to where you want to do this. That is going to override every anchor visited link everywhere. https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity

hutcheonb-moj: changed to class implementation, thanks Paul!

psoleckimoj (approved):
LGTM!

tpmcgowan:
Just thinking about this: the ticket was originally about removing visited link colour on 'Check your answers' page - which this PR does. However, there was subsequent confirmation on the ticket about also removing visited link colour across the application because it still appears in lots of places, e.g. search results, where it's not wanted as not really applicable.

We could therefore go and add the `govuk-link--no-visited-state` class in various other places, but I wonder whether adding something like:
```scss
.govuk-main-wrapper a:visited {
  color: $govuk-link-colour;
}
```
in `local.scss` for example would be preferable? It would then apply to the main body content on pages, but exclude header/footer. Interested in your thoughts, @psoleckimoj @hutcheonb-moj .

psoleckimoj:
@tpmcgowan can give it a try and make sure it works.

tpmcgowan:
Gave it a quick try and it seemed to work OK; just had to tweak it to not catch buttons that render as links:
```scss
.govuk-main-wrapper a:not(.govuk-button):visited {
  color: $govuk-link-colour;
}
```
So, perhaps worth a shot as an alternative to putting `govuk-link--no-visited-state` on all links (and remembering to do so for future ones!) ?

PR #328, opened 2022-09-21
VB-1212: Enable overbooking

tpmcgowan:
This change removes the restriction that prevents visit slots shown as 'Fully booked' from being chosen. The field is no longer disabled and instead additional text highlights how overbooked the slot is: e.g. "Fully booked (32 of 30 tables booked)".

Change is necessary at least temporarily because migrated data can exceed capacity.


server/@types/bapv.d.ts
```diff
@@ -139,6 +139,7 @@ export type VisitSlot = {
   startTimestamp: string
   endTimestamp: string
   availableTables: number
+  capacity: number
```
tpmcgowan, line 142: Adding capacity to `visitSlot` as it wasn't actually available on the date/time page - this results in most of the changes in this PR because of updates to test data, etc.

server/middleware/sessionCheckMiddleware.ts
```diff
@@ -30,7 +30,7 @@ export default function sessionCheckMiddleware({ stage }: { stage: number }): Re
       stage > 2 &&
       (!visitSessionData.visit ||
         !visitSessionData.visit.id ||
-        !visitSessionData.visit.availableTables ||
+        typeof visitSessionData.visit.availableTables === 'undefined' ||
```
tpmcgowan, line 33: Took a while to figure out this was why it would work with availability of, say, `-1` but not `0` 🤦‍♂️ !

server/views/pages/bookAVisit/dateAndTime.njk
```diff
@@ -22,7 +22,7 @@
       {% if (doubleBooked and not checked) -%}
         Prisoner has a visit
       {%- elseif slot.availableTables <= 0 -%}
-        Fully booked
+        Fully booked ({{ slot.capacity + (slot.availableTables | abs) }} of {{ slot.capacity}} tables booked)
```
psoleckimoj, line 25: If i was being really picky, I'd say you needed some code around the `s` in `tables` for when it's a single table available, but that can be done whenever as it's not going to happen very often, if ever 😁

psoleckimoj (dismissed):
LGTM!

server/views/pages/bookAVisit/dateAndTime.njk
tpmcgowan: Fixed!

psoleckimoj (approved):
LGTM!

PR #331, opened 2022-09-22
VB-1217: Move HEI into Audit function for prisonerSearch

psoleckimoj:
Not really a POC, just the first Audit function to be changed, if we're OK with this then I can do the rest in one PR.


tpmcgowan (approved):
LGTM!

PR #346, opened 2022-09-26
VB-1208: Update visit scheduler types

tpmcgowan:
Update types to take into account new endpoints and changes to DTOs. First step of switching over to use new endpoints.


psoleckimoj (approved):
LGTM!

PR #347, opened 2022-09-26
VB-1180: Give accordions unique ID to stop session based selection

psoleckimoj:
There is no way to stop the accordion component using sessions to remember what was open and closed, it overrides the values of `expanded` as that is already set correctly but ignored if an option was previously opened.

ID values can be 512Kb long, used uuid to ensure uniqueness.

Bad Paul, but the `docker-compose` having a common name of `redis` is causing no end of issues for me locally with other projects doing the same.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,11 +1,11 @@
 version: '3.1'
 services:
 
-  redis:
+  redis-bapv:
     image: 'bitnami/redis:5.0'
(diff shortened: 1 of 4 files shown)


hutcheonb-moj (approved):
LGTM

tpmcgowan (approved):
LGTM - nice solution!

PR #348, opened 2022-09-26
Update config for ts-jest v29

tpmcgowan:
Recent update of `ts-jest` from 28 to 29 broke recent speed improvements for tests. Issue seems to be deprecation of way config is specified (https://github.com/kulshekhar/ts-jest/blob/main/CHANGELOG.md#2900-2022-09-08).

This moves config to the new syntax (https://kulshekhar.github.io/ts-jest/docs/getting-started/options).

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,10 +36,12 @@
   },
   "jest": {
     "preset": "ts-jest",
-    "globals": {
-      "ts-jest": {
-        "isolatedModules": true
-      }
+    "transform": {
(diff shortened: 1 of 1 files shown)


psoleckimoj (approved):
LGTM!

Thanks @tpmcgowan , missed that one!

tpmcgowan:
I only spotted it as I couldn't figure out why my tests were all going slow again!

PR #350, opened 2022-09-27
VB-1208: Move booking journey to use new endpoints

tpmcgowan:
Remove use of the deprecated visit scheduler endpoints:
* `POST /visits` (for reserving a visit)
* `PUT /visits/{reference}` (for changing and booking a visit)

and replace them with support for the new endpoints:

* `POST /visits/slot/reserve` (make initial reservation)
* `PUT /visits/{applicationReference}/slot/change` (for making changes whilst visit still RESERVED)
* `PUT /visits/{applicationReference}/book` (for changing status from RESERVED to BOOKED)

This is **only** to support the booking journey so far; next step is the amend visit journey.


server/@types/bapv.d.ts
```diff
@@ -194,6 +194,7 @@ export type VisitSessionData = {
     phoneNumber: string
     contactName?: string
   }
+  applicationReference?: string
```
tpmcgowan, line 197: New endpoints rely on `applicationReference` rather than just `visitReference`.

server/routes/visit.test.ts
```diff
@@ -1654,153 +1654,6 @@ describe('/visit/:reference/update/select-date-and-time', () => {
         })
     })
   })
-
-  describe.skip('POST /visit/:reference/update/select-date-and-time', () => {
```
tpmcgowan, line 1658: Removing these blocks of tests (they were 'skipped' anyway) for now but will add coverage for update journey separately, likely as part of VB-1204.

server/data/visitSchedulerApiClient.ts
```diff
@@ -100,41 +100,42 @@ class VisitSchedulerApiClient {
...
             visitContact: visitor.personId === mainContactId,
           }
         }),
-        visitorSupport: visitData.visitorSupport,
+        visitorSupport: visitSessionData.visitorSupport,
       },
     })
   }
 
+  bookVisit(visitSessionData: VisitSessionData): Promise<Visit> {
```
psoleckimoj, line 135: Is there any point passing all the data to this method when it only uses the reference?

tpmcgowan: Good point 🤦‍♂️ ! I'll tweak it...

psoleckimoj (approved):
LGTM! Only one comment but doesn't stop anything from working.

server/data/visitSchedulerApiClient.ts
tpmcgowan: OK, changed to just work with `application` reference - which is all it needs.

psoleckimoj (approved):
LGTM!

PR #353, opened 2022-09-28
vb-1236, added application reference to audit service,

hutcheonb-moj:
also added application reference to the reserve and book calls to the audit service

## Description
Updated the audit service to accept application reference when making a booking or reservation
Included application reference in the audit service calls for both reserve and book


psoleckimoj (approved):
LGTM!

PR #354, opened 2022-09-29
VB-1208: Add support for new change booked visit endpoint

tpmcgowan:
Add support for `PUT /visits/{reference}/change`, which returns a visit with status set to either `RESERVED` (where date/time or restriction different) or `CHANGING` (any other changes) and with a new `applicationReference`.

### DIFF
--- a/server/data/visitSchedulerApiClient.ts
+++ b/server/data/visitSchedulerApiClient.ts
@@ -6,7 +6,7 @@ import {
   VisitSession,
   OutcomeDto,
   ReserveVisitSlotDto,
-  ChangeReservedVisitSlotRequestDto,
+  ChangeVisitSlotRequestDto,
 } from './visitSchedulerApiTypes'
 import { VisitSessionData } from '../@types/bapv'
 import config from '../config'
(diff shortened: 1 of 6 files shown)


server/data/visitSchedulerApiClient.ts
```diff
@@ -144,4 +158,19 @@ class VisitSchedulerApiClient {
   }
 }
 
+function convertMainContactToVisitContact(mainContact: VisitSessionData['mainContact']): {
```
psoleckimoj, line 161: Not sure if this would be better as a method inside the class? Just keeps it tidy?

psoleckimoj (approved):
LGTM! Only one comment but doesn't stop the merge.

server/data/visitSchedulerApiClient.ts
tpmcgowan: Good spot! Have tidied this up 🙂

psoleckimoj (approved):
LGTM!

server/data/visitSchedulerApiClient.ts
psoleckimoj: Approved 🙂

PR #355, opened 2022-09-29
vb-1024, additional support requests using journey loop

hutcheonb-moj:
## Description
Additional Support tests now use the journey loop, to run through both the book and amend journeys

### DIFF
--- a/server/routes/visitJourney/additionalSupport.test.ts
+++ b/server/routes/visitJourney/additionalSupport.test.ts
@@ -12,6 +12,12 @@ const systemToken = async (user: string): Promise<string> => `${user}-token-1`
 let flashData: Record<'errors' | 'formValues', Record<string, string | string[]>[]>
 let visitSessionData: VisitSessionData
 
+// run tests for booking and update journeys
+const testJourneys = [
+  { urlPrefix: '/book-a-visit', previousVisitReference: undefined },
+  { urlPrefix: '/visit/ab-cd-ef-gh/update', previousVisitReference: 'ab-cd-ef-gh' },
+]
+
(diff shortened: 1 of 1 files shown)


server/routes/visitJourney/additionalSupport.test.ts
```diff
@@ -46,379 +52,385 @@ afterEach(() => {
...
+          expect($('[data-test="WHEELCHAIR"]').prop('checked')).toBe(true)
+          expect($('[data-test="OTHER"]').prop('checked')).toBe(true)
+          expect($('#otherSupportDetails-error').text()).toContain('Enter details of the request')
+          expect(flashProvider).toHaveBeenCalledWith('errors')
+          expect(flashProvider).toHaveBeenCalledWith('formValues')
+          expect(flashProvider).toHaveBeenCalledTimes(2)
+        })
+    })
   })
 })
```
psoleckimoj, line 233: Not sure you need to end thee loop and restart on 235? Could just remove 233 and 235? Not tried but should work.

tpmcgowan:
I think tests have failed again because of the update journey feature flag issue.

hutcheonb-moj:
remembered the flag as soon as I changed it to a full PR, working on it now, thanks!

server/routes/visitJourney/additionalSupport.test.ts
hutcheonb-moj: I had tried it without and found the grouping was strange, will re-attempt though while I add the flag

tpmcgowan (approved):
LGTM!

PR #357, opened 2022-09-30
vb-1024, main contact tests updated

hutcheonb-moj:
## Description
Unit tests for main contact now test both journeys, also removed tests from visit relating to main contact

Linter pointed out some now unused code, so removed that from visit.test


psoleckimoj (approved):
LGTM!

PR #362, opened 2022-10-04
VB-1116: Update visit routes - add session and URL validation

tpmcgowan:
This chang:
* adds the existing `sessionCheckMiddleware` to the update visit routes (works without change now that the journey is effectively the same for book and update)
* adds an additional check to ensure that the visit reference in the URL (if present) matches that in the `visitSessionData`


psoleckimoj (approved):
LGTM!

PR #363, opened 2022-10-05
vb-1204, select visitor tests updated

hutcheonb-moj:
## Description
Moved the select visit tests out of the visit.test file
Select visitor tests now runs through both the booking and the update journey to fully test functionality


psoleckimoj (approved):
LGTM!

PR #364, opened 2022-10-05
vb-1254, remove cancel and update button on past visit

hutcheonb-moj:
## Description
Remove the cancel and update buttons on the visits summary page, when the visit start date has already passed

3 tests have been skipped, as the test data has tests in the past, these fail.
Attempted to update the test data but this would only be a temporary fix.

Tried using jest.useFakeTime but this caused timeouts on all of the tests. 

Needs more work but having no luck getting these tests working correctly

### DIFF
--- a/server/routes/visit.test.ts
+++ b/server/routes/visit.test.ts
@@ -177,7 +177,7 @@ describe('GET /visit/:reference', () => {
     })
   })
 
-  it('should render full booking summary page with prisoner, visit and visitor details, with default back link', () => {
+  it.skip('should render full booking summary page with prisoner, visit and visitor details, with default back link', () => {
     return request(app)
       .get('/visit/ab-cd-ef-gh')
       .expect(200)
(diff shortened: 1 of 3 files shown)


tpmcgowan:
To get the tests working, I think you need to include (in relevant `beforeEach`):
```javascript
const fakeDate = new Date('2022-07-11T09:00:00.000Z')
jest.useFakeTimers({ advanceTimers: true, now: fakeDate })
```

and then an `afterEach` / `afterAll` at an appropriate place with `jest.useRealTimers()` to reset it.

Key to getting it working is the `advanceTimers` config I think because we've got `async` stuff going on.

server/routes/visit.test.ts
```diff
@@ -159,6 +159,9 @@ describe('GET /visit/:reference', () => {
       modifiedTimestamp: '2022-02-14T10:05:00',
     }
 
+    const fakeDate = new Date('2022-01-01')
+    jest.useFakeTimers({ advanceTimers: true, now: fakeDate })
```
tpmcgowan, line 163: I've looked into this a bit further and, instead of my original `advanceTimers` suggestion, you could alternatively do:
```javascript
jest.useFakeTimers({ doNotFake: ['nextTick'], now: fakeDate })
```
This fixes the problem because it's the faking of [`nextTick` ](https://nodejs.org/en/docs/guides/event-loop-timers-and-nexttick/) that causes it to hang. 

`advanceTimers` automatically keeps it running, but we could just not fake it in the first place 🤷‍♂️ ! (I think either is fine...)

tpmcgowan (approved):
LGTM!

PR #367, opened 2022-10-06
VB-1251: Ensure original slot can be re-selected when updating a visit

tpmcgowan:
Previously, during an update if the slot was changed and the user went back (before confirming the updated booking) then they would be unable to re-select the original slot. This change fixes that by keeping track of the original visit slot so the relevant date/time checkbox isn't disabled.

Outstanding issues:
1. needs some more tests - coming next!
2. Awaiting confirmation of text changes to better label original/changed slots

Works without these two caveats though so putting up for review so as not to block other tickets that will be affected by these changes (e.g. VB-1079)

### DIFF
--- a/server/@types/bapv.d.ts
+++ b/server/@types/bapv.d.ts
@@ -184,8 +184,8 @@ export type VisitSessionData = {
     previousRestrictions?: OffenderRestriction[]
   }
   visit?: VisitSlot
+  originalVisitSlot?: VisitSlot
   visitRestriction?: Visit['visitRestriction']
-  previousVisitRestriction?: Visit['visitRestriction']
   closedVisitReason?: 'prisoner' | 'visitor'
   visitors?: VisitorListItem[]
   visitorSupport?: VisitorSupport[]
--- a/server/views/pages/bookAVisit/dateAndTime.njk
+++ b/server/views/pages/bookAVisit/dateAndTime.njk
@@ -15,10 +15,11 @@
 {% macro buildSlotsRadioItems(slotsList, radios) %}
   {% for slot in slotsList %}
 
-    {%- set doubleBooked = slot.sessionConflicts and "DOUBLE_BOOKED" in slot.sessionConflicts %}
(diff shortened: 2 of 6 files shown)


server/views/pages/bookAVisit/dateAndTime.njk
```diff
@@ -15,10 +15,11 @@
 {% macro buildSlotsRadioItems(slotsList, radios) %}
   {% for slot in slotsList %}
 
-    {%- set doubleBooked = slot.sessionConflicts and "DOUBLE_BOOKED" in slot.sessionConflicts %}
+    {%- set doubleBooked = slot.id != originalSelectedSlot.id and (slot.sessionConflicts and "DOUBLE_BOOKED" in slot.sessionConflicts) %}
     {%- set checked = true if formValues['visit-date-and-time'] and formValues['visit-date-and-time'] == slot.id else false %}
 
     {%- set tableText %}
+      {# {% if (slot.id == originalSelectedSlot.id) -%}ORIGINAL SLOT {% endif %} #}
```
tpmcgowan, line 22: Commented out while awaiting what the text should be to indicate that this is the original slot.

server/@types/bapv.d.ts
```diff
@@ -184,8 +184,8 @@ export type VisitSessionData = {
     previousRestrictions?: OffenderRestriction[]
   }
   visit?: VisitSlot
+  originalVisitSlot?: VisitSlot
```
tpmcgowan, line 187: Removed `previousVisitRestriction` as storing `originalVisitSlot` includes this.

(Side note - just wondering if `visit` here would be better named as `visitSlot` because that's what it actually is, rather than the visit itself! Any thoughts, @psoleckimoj ?)

psoleckimoj: @tpmcgowan sounds good to me 👍

psoleckimoj: Less confusion elsewhere 🙂

tpmcgowan: 👍 will rename that then - but as a separate PR

psoleckimoj (approved):
LGTM!

PR #371, opened 2022-10-07
Tweak jest code coverage config

tpmcgowan:
Small tweak to Jest configuration so that code coverage runs when doing full test suite but not automatically when just running an individual test (e.g. with `npx jest <file.test.ts>`).


psoleckimoj (approved):
LGTM!

PR #373, opened 2022-10-07
VB-1251: Improve slot labelling for book/update

tpmcgowan:
Improve the labelling of date/time slots so that rather than displaying table count or "fully booked" when there is already a reserved slot (either when revising an initial booking or amending a booked visit) a more descriptive label is used.

### DIFF
--- a/server/views/pages/bookAVisit/dateAndTime.njk
+++ b/server/views/pages/bookAVisit/dateAndTime.njk
@@ -19,8 +19,11 @@
     {%- set checked = true if formValues['visit-date-and-time'] and formValues['visit-date-and-time'] == slot.id else false %}
 
     {%- set tableText %}
-      {# {% if (slot.id == originalSelectedSlot.id) -%}ORIGINAL SLOT {% endif %} #}
-      {% if (doubleBooked and not checked) -%}
+      {% if (slot.id == originalSelectedSlot.id) -%}
+        Time slot reserved by the original booking
+      {% elseif (checked and not originalSelectedSlot) -%}
+        Time slot reserved for this booking
+      {% elseif (doubleBooked and not checked) -%}
         Prisoner has a visit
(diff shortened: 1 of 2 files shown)


server/views/pages/bookAVisit/dateAndTime.njk
```diff
@@ -19,8 +19,11 @@
     {%- set checked = true if formValues['visit-date-and-time'] and formValues['visit-date-and-time'] == slot.id else false %}
 
     {%- set tableText %}
-      {# {% if (slot.id == originalSelectedSlot.id) -%}ORIGINAL SLOT {% endif %} #}
-      {% if (doubleBooked and not checked) -%}
+      {% if (slot.id == originalSelectedSlot.id) -%}
```
tpmcgowan, line 22: The logic in all these `if`/`elseif` getting a bit complicated but couldn't think of a better way to structure it 🤷‍♂️ . (Hoping the accompanying tests help explain!)

hutcheonb-moj (approved):
LGTM

server/views/pages/bookAVisit/dateAndTime.njk
psoleckimoj: Can you do `switch` in Nunjucks, that's the only other way to do it really but I've no preference. Other option is a function that takes the relevant args and returns a message.

psoleckimoj (approved):
LGTM!

server/views/pages/bookAVisit/dateAndTime.njk
tpmcgowan: Thanks for the suggestions. Doesn't look like Nunjucks has `switch`. Going to leave it as is given that this date/time picker is being re-designed - an opportunity then to tidy up some of this kind of stuff.

PR #375, opened 2022-10-07
vb-1079, formatting matches for both journeys

hutcheonb-moj:
## Description
Formatting now matches for both the update and booking journeys for the date and time restriction messages

Tests to be completed for the date and time routes test

### DIFF
--- a/server/views/pages/bookAVisit/dateAndTime.njk
+++ b/server/views/pages/bookAVisit/dateAndTime.njk
@@ -54,12 +54,24 @@
     </p>
 
     {% if restrictionChangeMessage %}
+
+      {% set restrictionChangeReasonHtml %}
+          {{ govukWarningText({
+            classes: "govuk-!-margin-bottom-0",
+            html: '<span class="govuk-!-font-size-24">' + restrictionChangeMessage + '</span>',
+            iconFallbackText: "Warning",
+            attributes: {
+              "data-test": "restriction-change-reason"
+          }
+        }) }}
+      {% endset %}
       {{ mojBanner({
-        type: 'information',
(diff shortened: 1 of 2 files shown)


server/routes/visitJourney/dateAndTime.test.ts
```diff
@@ -91,6 +98,15 @@ const slotsList: VisitSlotList = {
             visitRoomName: 'room name',
             visitRestriction: 'OPEN',
           },
+          {
+            id: '5',
```
tpmcgowan, line 102: There's already an `id` with the value `5` within this `slotsList` so worth renumbering to remove duplicate - otherwise we might get unexpected results

server/routes/visitJourney/dateAndTime.test.ts
```diff
@@ -481,3 +490,147 @@ testJourneys.forEach(journey => {
     })
   })
 })
+
+describe(`Update journey specific testing`, () => {
+  it('Restriction change: Open to closed, no visit slots available - visitor', () => {
```
tpmcgowan, line 495: Very minor point on wording of test names (this and the other ones): I think rather than "no visit slots available" it's more accurate to say something like "no availability for the selected visit slot".

The difference is that we're using "visit slot" to mean distinct time and dates where there may (or may not!) be availability to schedule a visit.

tpmcgowan (commented):
Looks good - just a comment about making sure `id`s are unique in the test data (and minor observation on naming )

tpmcgowan (dismissed):
LGTM!

tpmcgowan (approved):

PR #378, opened 2022-10-07
VB-1268: Trim text input on prisoner search

tpmcgowan:
Trim text input when searching for a prisoner.

(While looking in Application Insights, I noticed that users were commonly submitting search queries with trailing spaces so there was lots of unnecessary `%20` type encodings in the logs.)


psoleckimoj (approved):
lgtm

PR #379, opened 2022-10-11
VB 1255, cancelled visit message on summary page

hutcheonb-moj:
## Description
Visit cancelled message on visit summary page now uses GDS warning component.

Tests run for visitor cancelled and administrative cancellation as they use different wordings.

Had to include some CSS as the GDS component doesn't natively support multiple lines of text

### DIFF
--- a/assets/sass/local.scss
+++ b/assets/sass/local.scss
@@ -253,3 +253,12 @@
     margin-bottom: govuk-spacing(3);
   }
 }
+
+.moj-banner__message {
+  .app-warning-text--long {
+    margin-bottom: 0;
+    span.govuk-warning-text__icon {
+      margin-top: 5px;
+    }
+  }
+}
--- a/assets/sass/local.scss
+++ b/assets/sass/local.scss
@@ -253,12 +253,3 @@
     margin-bottom: govuk-spacing(3);
   }
 }
-
-.moj-banner__message {
-  .app-warning-text--long {
-    margin-bottom: 0;
-    // span.govuk-warning-text__icon {
-    //   margin-top: 5px;
-    // }
-  }
-}
--- a/assets/sass/local.scss
+++ b/assets/sass/local.scss
@@ -257,8 +257,8 @@
 .moj-banner__message {
   .app-warning-text--long {
     margin-bottom: 0;
-    span.govuk-warning-text__icon {
-      margin-top: 5px;
-    }
+    // span.govuk-warning-text__icon {
+    //   margin-top: 5px;
+    // }
   }
 }
--- a/server/views/pages/visit/summary.njk
+++ b/server/views/pages/visit/summary.njk
@@ -29,16 +29,16 @@
 
                 {% set cancelledBy %}
                   {% if status[0] === 'ADMINISTRATIVE' %}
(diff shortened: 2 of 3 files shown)


server/views/pages/visit/summary.njk
```diff
@@ -20,6 +22,39 @@
...
       <h1 class="govuk-heading-l">{{ pageHeaderTitle }}</h1>
 
+       {% if visit.visitStatus === "CANCELLED" %}
+        {% set status = visit.outcomeStatus.split("_") %}
+            {%- for visitNote in visit.visitNotes %}
+              {% if visitNote.type == "VISIT_OUTCOMES" %}
+
+                {% set cancelledBy %}
+                  {% if status[0] === 'ADMINISTRATIVE' %}
+                    {{ 'because of an administrative error.' }}
```
tpmcgowan, line 32: Cleverly split out! Just wondering though about the text ("because of an administrative error") used? Couldn't see it on the ticket so just checking if Ed/Phoebe came up with this. If not, might be worth checking with them if they're happy with what the text will end up being under the [four possible cases](https://github.com/ministryofjustice/book-a-prison-visit-staff-ui/blob/main/server/constants/visitCancellationReasons.ts).

Only suggesting this as I wondered if something like 'administrative reasons' might be a bit more generic than 'administrative error' 🤷‍♂️

server/views/pages/visit/summary.njk
```diff
@@ -20,6 +22,39 @@
...
+                  {% else %}
+                    {{ 'by the ' + status[0] | lower  + '.' }}
+                  {% endif %}
+                {% endset %}
+
+                {% set cancelledVisitHtml %}
+                  {{ govukWarningText({
+                    classes: "app-warning-text--long govuk-!-padding-left-4",
+                    html: "<span>This visit was cancelled " + cancelledBy +"</span><br><span>Reason: " + visitNote.text + ".</span>",
```
tpmcgowan, line 41: I would remove the `.` at the end where you have `".</span>"` - otherwise you end up with two if the user entered a full stop themselves!

assets/sass/local.scss
```diff
@@ -253,3 +253,12 @@
...
 }
+
+.moj-banner__message {
+  .app-warning-text--long {
+    margin-bottom: 0;
+    span.govuk-warning-text__icon {
+      margin-top: 5px;
+    }
+  }
+}
```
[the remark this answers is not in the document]
hutcheonb-moj: I thought it looked horrible the way it was without including the margin top, as that component is made for aligning a single line. Completely missed the margin bottom rule though thanks!

server/views/pages/visit/summary.njk
hutcheonb-moj: I noticed that the wording would have been 'by the administrative error' if I implemented this case like I have the others, was hoping to get some feedback on what was best for this case, but knew we would deal with it differently

server/views/pages/visit/summary.njk
hutcheonb-moj: hadn't thought of that!

tpmcgowan (approved):
LGTM!

PR #382, opened 2022-10-12
Change to CSP to fix for Safari on localhost

tpmcgowan:
This change removes the `'upgrade-insecure-requests'` directive from the CSP when the application is in `development` mode.

This fixes the application for Safari when running on `localhost`. Other browsers don't try to upgrade insecure requests on `localhost` but Safari does - which means that assets like CSS, etc are blocked.

(This issue arose when Helmet was upgraded from [version 4 to 5](https://github.com/helmetjs/helmet/blob/main/CHANGELOG.md#500---2022-01-02) and a new default CSP was used that included `upgrade-insecure-requests`.)


psoleckimoj (dismissed):
LGTM!

hutcheonb-moj (approved):
LGTM

PR #383, opened 2022-10-12
VB-1223, changed beta banner to support smaller view

hutcheonb-moj:
## Description
Restructured the beta banner section, to stop it sitting offset on the second line, when on a small view


tpmcgowan (approved):
LGTM!

PR #384, opened 2022-10-13
VB-1223: Add padding to right of beta banner

tpmcgowan:
Stop the beta banner text getting too close to the right-hand edge of the screen on smaller sizes.


AledWEvans (commented):
LGTM

AledWEvans (approved):
LGTM

PR #386, opened 2022-10-17
VB-1280: Enable update journey in prod

tpmcgowan:
Turn on the feature flag for update journey for prod so it's ready for upcoming release.


psoleckimoj (approved):
LGTM!

PR #387, opened 2022-10-18
VB-1325: Update AppInsights connection string

psoleckimoj:
Use the correct connection string, copied from template app https://github.com/ministryofjustice/hmpps-template-typescript/blob/main/server/utils/azureAppInsights.ts


tpmcgowan (approved):
LGTM

PR #391, opened 2022-10-21
VB-729: Booking journey Cypress tests

tpmcgowan:
* Refactor of integration test Wiremock stubs to reflect API changes
* Use common test objects from `TestData.ts`
* Add test for Home page tiles
* Add full booking journey test
* Update existing prisoner search tests
* Split off Prisoner profile page tests


.circleci/config.yml
```diff
@@ -119,7 +119,7 @@ jobs:
           command: sudo apt-get install libxss1
       - run:
           name: Get wiremock
-          command: curl -o wiremock.jar https://repo1.maven.org/maven2/com/github/tomakehurst/wiremock-standalone/2.27.1/wiremock-standalone-2.27.1.jar
+          command: curl -o wiremock.jar https://repo1.maven.org/maven2/com/github/tomakehurst/wiremock-jre8-standalone/2.35.0/wiremock-jre8-standalone-2.35.0.jar
```
tpmcgowan, line 122: Upgrading to latest Wiremock version so it matches what runs locally in Docker from the `wiremock/wiremock` image. Older version here was causing builds in CircleCI to fail.

psoleckimoj: Is it possible to fix the wiremock image to a version? So we can upgrade when we choose?

psoleckimoj (approved):
LGTM!

.circleci/config.yml
tpmcgowan: Good suggestion! I've set it to the same `2.35.0` tag in `docker-compose-test.yml` and commented in both files as a reminder that the versions should match.

psoleckimoj (approved):
LGTM!

PR #398, opened 2022-10-26
VB-1321: Example print page

psoleckimoj:


tpmcgowan:
Figured we can probably close this as it's been sat here over a year and doesn't seem to be a requirement any more 🤷‍♂️

PR #399, opened 2022-11-02
VB-1202, Age now being displayed on Select Visitor and Booking Details

hutcheonb-moj:
## Description
Remove Child/Adult from the select visitor and the booking summary pages, replaced this with the age of the visitor.
Added a new nunjucks filter 'displayAge' which returns a 'grammatically correct' age based on current date - date given


psoleckimoj:
@hutcheonb-moj looks good but could do with some test to check the month age thing and maybe one for 1 year and 1 month to check the plurals?

hutcheonb-moj:
@psoleckimoj up now, forgot to push after a particularly slow commit...

psoleckimoj:
@hutcheonb-moj can't see any extra tests I suggested?

server/utils/nunjucksSetup.ts
```diff
@@ -84,6 +84,29 @@ export function registerNunjucks(app?: express.Express): Environment {
...
+    const dob = new Date(dateOfBirth)
+    const years = today.getFullYear() - dob.getFullYear()
+    if (years > 1) {
+      ageString = `${years} years old`
+    } else if (years === 1) {
+      ageString = '1 year old'
+    } else if (years < 1) {
+      const months = today.getMonth() - dob.getMonth()
+
+      if (months > 1) {
```
tpmcgowan, line 100: Perhaps very unlikely to occur, but if a baby less than one month old then I guess this would show `0 month old` but should it be `0 months`?

hutcheonb-moj: Just re-checked and realised that was mentioned on the ticket, will go back and add it now, Thanks!

hutcheonb-moj (commented):
@psoleckimoj did you not mean these?
https://github.com/ministryofjustice/book-a-prison-visit-staff-ui/pull/399/commits/153b3fc14b83691b542db83cdced7d8a40387860

server/utils/nunjucksSetup.ts
tpmcgowan: That looks like it should do the trick. For readability, could consider flipping the condition so instead of:
```if (months > 1 || months === 0) ...```
 it's
```if (months === 1) { // use singular } else { // use plural }``` 

Could also cover these extra cases in `nunjucksSetup.test.ts`

tpmcgowan (approved):
LGTM!

PR #404, opened 2022-11-04
VB-1390: Use selected establishment rather than hard-coded Hewell

tpmcgowan:
Part of work to remove hard-coded references to `Hewell / HEI` and use the `selectedEstablishment` value in the session. More still to be done (separately, to keep this PR manageable).

Some further tidying (noted in code comments) will be necessary once other tickets are complete.


psoleckimoj (approved):
LGTM!

PR #406, opened 2022-11-07
VB-1390: Use selected establishment rather than hard-coded 'HEI'

tpmcgowan:
Further to #404, remove remaining hard-coded `Hewell / HEI` references and use `selectedEstablishment` in the session instead.


psoleckimoj (approved):
LGTM!

PR #407, opened 2022-11-07
VB-1495: Add default background colour to date picker

tpmcgowan:
Add a background colour to the pop-up date picker on 'View visits by date' page so that at narrow window size the main content isn't visible underneath.


hutcheonb-moj (approved):
LGTM

PR #408, opened 2022-11-08
VB-1409: Add type for 'Prison'

tpmcgowan:
Add type for `Prison` - using same property names as https://github.com/ministryofjustice/prison-register for future compatibility. Precursor to adding a `SupportedPrisonsService` into the application.

### DIFF
--- a/server/routes/changeEstablishment.ts
+++ b/server/routes/changeEstablishment.ts
@@ -1,18 +1,18 @@
 import type { RequestHandler, Router } from 'express'
+import { Prison } from '../@types/bapv'
 
 import asyncMiddleware from '../middleware/asyncMiddleware'
 
 export default function routes(router: Router): Router {
   const get = (path: string | string[], handler: RequestHandler) => router.get(path, asyncMiddleware(handler))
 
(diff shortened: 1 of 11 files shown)


server/routes/changeEstablishment.ts
```diff
@@ -1,18 +1,18 @@
...
-    { prisonId: 'HEI', name: 'Hewell (HMP)' },
-    { prisonId: 'BLI', name: 'Bristol (HMP)' },
+  const enabledPrisons: Prison[] = [
+    { prisonId: 'HEI', prisonName: 'Hewell (HMP)' },
+    { prisonId: 'BLI', prisonName: 'Bristol (HMP)' },
   ]
 
   get('/', (req, res, next) => {
     res.render('pages/changeEstablishment', {
-      hidePhaseBanner: true,
```
psoleckimoj, line 15: Just checking this was meant as part of this PR?

tpmcgowan: Well spotted - yes, that was intentional as the change establishment page is supposed to have the phase banner 🙂

PR #410, opened 2022-11-10
VB-1412: Keep selected establishment when clearing session

tpmcgowan:
When `clearSession` is called, keep the `selectedEstablishment` set. This will prevent it inadvertently being defaulted to a different value to the one a user may have chosen.


psoleckimoj (approved):
LGTM!

PR #411, opened 2022-11-10
VB-1388: Add feature flag for establishment switcher

tpmcgowan:
Adds feature flag for establishment switcher, specifically so far the `/change-establishment` route. Initially only enabled on the `dev` environment. Getting this in prior to merging #404, which will actually allow the establishment to be changed.

(Tests will be added to #404 once this is PR is in.)


psoleckimoj (approved):
LGTM!

PR #412, opened 2022-11-11
VB-1401: Prisoner profile page - show correct prison on upcoming/past visits tab

tpmcgowan:
On the prisoner profile page, the visit location on the upcoming and past visits tabs was hard-coded to Hewell. This change looks up the location for each visit using the `SupportedPrisonsService`.

* add `getSupportedPrison(prisonId: string)` to service to get details of a specific prison
* use this when generating the list of visit items on the profile page tabs


server/services/supportedPrisonsService.ts
```diff
@@ -10,6 +10,10 @@ export default class SupportedPrisonsService {
     private readonly systemToken: SystemToken,
   ) {}
 
+  async getSupportedPrison(prisonId: string): Promise<Prison> {
```
tpmcgowan, line 13: Doesn't actually need to be `async` at the moment but figured it will be once we have more prisons and move from static file of values to looking them up in API.

server/app.ts
```diff
@@ -83,6 +83,7 @@ export default function createApp(userService: UserService): express.Application
         prisonApiClientBuilder,
         visitSchedulerApiClientBuilder,
         prisonerContactRegistryApiClientBuilder,
+        new SupportedPrisonsService(visitSchedulerApiClientBuilder, systemToken),
```
tpmcgowan, line 86: Does passing a new instance of the service in here seem OK? So far other services we have are just taking API client builders, not other services... Tried a few different ways to get this in at various places but this seemed to work easiest for testing.

psoleckimoj: If it works the same as other services, can it not just use a builder like the others as they're testable?

tpmcgowan: Not sure I made my question clear. Will try and explain better...!

Our existing routes have various services passed to them and these services themselves (up to now) only take various `...ApiClientBuilder`'s.

I've added another **service** (`SupportedPrisonsService`) as a dependency for `PrisonerProfileService`. Thus far our other services have only had various API clients as dependencies (via 'builders'); not other services.

The other services don't have builders so I wasn't sure if that was the way to go; so I just passed a new instance of the `SupportedPrisonsService` into the `PrisonerProfileService` when it itself is created to pass to the route. So, I was wondering if this seemed OK?

Having looked around, I think more fundamentally we should probably incorporate https://github.com/ministryofjustice/hmpps-template-typescript/pull/106 from the template app as then there wouldn't need to be so much going on in `app.ts` and so many changes rippling through when a new service or API client is added? E.g. there's nice examples of this in [create-and-vary-a-licence](https://github.com/ministryofjustice/create-and-vary-a-licence/blob/main/server/services/index.ts) and [hmpps-welcome-people-into-prison-ui](https://github.com/ministryofjustice/hmpps-welcome-people-into-prison-ui/blob/main/server/services/index.ts)

psoleckimoj (approved):
LGTM!

PR #413, opened 2022-11-11
VB-1402: Booking details page - show location if a supported prison

tpmcgowan:
Prisoner's current location currently showing as either Hewell or Unknown on booking details page. 

This change displays the prisoners current location as long as it is in the list of supported prisons. Otherwise 'Unknown' is still shown (as a precaution against, for example, revealing a transfer or similar).


psoleckimoj (approved):
LGTM!

PR #414, opened 2022-11-11
VB-1400, show change establishment link and tests

hutcheonb-moj:
## Description
Added change establishment link to Index, search pages and results
Added tests to ensure link only being displayed when feature flag is enabled

### DIFF
--- a/assets/sass/local.scss
+++ b/assets/sass/local.scss
@@ -1,7 +1,9 @@
 .govuk-main-wrapper a:not(.govuk-button, .govuk-tabs__tab):visited {
   color: $govuk-link-colour;
 }
-
+.location-bar .location-bar__location a:visited {
+  color: $govuk-link-colour;
+}
 .govuk-main-wrapper {
   min-height: 600px;
 }
--- a/assets/sass/local.scss
+++ b/assets/sass/local.scss
@@ -1,6 +1,4 @@
-.govuk-main-wrapper a:not(.govuk-button, .govuk-tabs__tab):visited {
-  color: $govuk-link-colour;
-}
+.govuk-main-wrapper a:not(.govuk-button, .govuk-tabs__tab):visited,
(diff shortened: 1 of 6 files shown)


server/routes/index.test.ts
```diff
@@ -25,6 +27,19 @@ describe('GET /', () => {
         expect(res.text).toContain('Book a visit')
         expect(res.text).toContain('Change a visit')
         expect(res.text).toContain('View visits by date')
+        expect(res.text).toContain('Change establishment')
```
tpmcgowan, line 30: All looks good - I'd just be tempted to take advantage of the `data-test` attribute (`data-test="change-establishment"`) you added for this element to make checking for this a bit more targeted  here and in the other tests in this PR. Then you can be sure you've found this element rather than happening to find a match anywhere on the page.

psoleckimoj: ☝️ 😄

assets/sass/local.scss
```diff
@@ -1,7 +1,9 @@
 .govuk-main-wrapper a:not(.govuk-button, .govuk-tabs__tab):visited {
   color: $govuk-link-colour;
 }
-
+.location-bar .location-bar__location a:visited {
```
tpmcgowan, line 4: Very minor point - just occurred to me that you could combine these two rules as they're doing the same thing; so something like:
```scss
.govuk-main-wrapper a:not(.govuk-button, .govuk-tabs__tab):visited,
.location-bar__location a:visited {
   color: $govuk-link-colour;
 }
```

tpmcgowan (approved):
LGTM!

assets/sass/local.scss
psoleckimoj: Was writing this as a suggestion, you got to it before me 😆

tpmcgowan (approved):
LGTM!

PR #415, opened 2022-11-15
VB-1409: Use Prison Register API to get list of all prisons

tpmcgowan:
This change adds support for the [Prison Register API](https://prison-register-dev.hmpps.service.justice.gov.uk/swagger-ui/index.html) to get a list of all prisons, which is used to find prison names for supported prisons.

There is further work to do to optimise this so the full details of supported prisons are stored in the session to avoid repeatedly calling the endpoint to get all prisons - splitting it up to keep PRs manageable.


integration_tests/mockApis/responses/allPrisons.ts
```diff
@@ -1,6 +1,4 @@
-import { Prison } from '../@types/bapv'
-
-const prisons: Prison[] = [
+export default [
```
tpmcgowan, line 1: Git seems to be suggesting I've moved this file - not quite what's happened 🙄! Have made this `responses` folder here to start trying to separate the test data from the stub API definitions.

server/data/__testutils/testObjects.ts
```diff
@@ -0,0 +1,16 @@
+import { Prison } from '../prisonRegisterApiTypes'
```
tpmcgowan, line 1: Made a start on a way to avoid duplicating lots of test data across the unit tests.

tpmcgowan:
Reverting back to draft as I'm going to re-jig the `Prison` type to make it simpler to work with and look at just caching the retrieved list of prisons within the `SupportedPrisonsService` class.

tpmcgowan:
Closing in favour of a slightly different approach in #422.

PR #416, opened 2022-11-15
VB-1400, add redirect to change establishment page

hutcheonb-moj:
## Description
Take request refer as a URL parameter when coming from any of the search pages.
If coming from index or when auto-directed here (VB-1412) then it will redirect back to index 

Currently no referral tests implemented, beyond the regular index.

Really just a PR for input

### DIFF
--- a/assets/sass/local.scss
+++ b/assets/sass/local.scss
@@ -2,6 +2,9 @@
 .location-bar .location-bar__location a:visited {
   color: $govuk-link-colour;
 }
+.location-bar .location-bar__location a {
+  font-weight: 400;
+}
 .govuk-main-wrapper {
   min-height: 600px;
 }
--- a/assets/sass/local.scss
+++ b/assets/sass/local.scss
@@ -2,9 +2,6 @@
 .location-bar .location-bar__location a:visited {
   color: $govuk-link-colour;
 }
-.location-bar .location-bar__location a {
-  font-weight: 400;
-}
 .govuk-main-wrapper {
   min-height: 600px;
 }
--- a/server/routes/changeEstablishment.ts
+++ b/server/routes/changeEstablishment.ts
@@ -35,10 +35,9 @@ export default function routes(router: Router, supportedPrisonsService: Supporte
     const supportedPrisons = await supportedPrisonsService.getSupportedPrisons(res.locals.user?.username)
     await body('establishment').isIn(getPrisonIds(supportedPrisons)).withMessage('No prison selected').run(req)
 
-    let referrer = (req.query?.referrer as string) ?? ''
-    if (referrer.length === 0) {
-      referrer = '/'
-    }
+    const referrer = (req.query?.referrer as string) ?? ''
+    const safeReturnUrl =
+      referrer.length === 0 || referrer.indexOf('://') > 0 || referrer.indexOf('//') === 0 ? '/' : referrer
 
     const errors = validationResult(req)
     if (!errors.isEmpty()) {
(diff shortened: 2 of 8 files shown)


server/routes/changeEstablishment.ts
```diff
@@ -22,16 +23,24 @@ export default function routes(router: Router, supportedPrisonsService: Supporte
   get('/', establishmentSwitcherCheckMiddleware, async (req, res) => {
     const supportedPrisons = await supportedPrisonsService.getSupportedPrisons(res.locals.user?.username)
 
+    const refer = (req.query?.refer as string) ?? ''
+
     res.render('pages/changeEstablishment', {
       errors: req.flash('errors'),
       supportedPrisons,
+      refer,
```
psoleckimoj, line 31: Better name would be `referrer` to be consistent with terminology 👍

psoleckimoj: Same for the querystring.

server/routes/search.ts
```diff
@@ -24,12 +24,19 @@ export default function routes(
     )
 
   get(['/prisoner', '/prisoner-visit'], (req, res) => {
+    const isVisit = req.originalUrl.includes('-visit')
     const search = req?.body?.search
-
+    let establishmentHref = ''
```
psoleckimoj, line 29: Better to default to `'/search/prisoner/'` and only change on the `if` condition

assets/sass/local.scss
```diff
@@ -2,6 +2,9 @@
 .location-bar .location-bar__location a:visited {
   color: $govuk-link-colour;
 }
+.location-bar .location-bar__location a {
+  font-weight: 400;
```
tpmcgowan, line 6: Rather than specifying a number for the weight, I'd be tempted to either:
* use `$govuk-font-weight-regular` here (see [Sass API Reference](https://frontend.design-system.service.gov.uk/sass-api-reference/#govuk-font-weight-regular)); or,
* instead add the `govuk-!-font-weight-regular` class in the template (see [Font override classes](https://design-system.service.gov.uk/styles/typography/#font-override-classes))

psoleckimoj:
@hutcheonb-moj apologies, all looks good, but wondering if adding some checking on the referrer might be worthwhile? Just to check they're only ever relative URLs? The Hub uses this:

```
const getSafeReturnUrl = ({ returnUrl = '/' } = {}) =>
  // type-check to mitigate "type confusion through parameter tampering", where an attacker
  // coerces the param to an array to bypass the indexOf checks
  typeof returnUrl !== 'string' ||
  returnUrl.indexOf('://') > 0 ||
  returnUrl.indexOf('//') === 0
    ? '/'
    : returnUrl;
```

server/routes/changeEstablishment.ts
```diff
@@ -22,16 +22,23 @@ export default function routes(router: Router, supportedPrisonsService: Supporte
   get('/', establishmentSwitcherCheckMiddleware, async (req, res) => {
     const supportedPrisons = await supportedPrisonsService.getSupportedPrisons(res.locals.user?.username)
 
+    const referrer = (req.query?.referrer as string) ?? ''
```
psoleckimoj, line 25: Check needs to happen anywhere you're pulling the referrer, stick it as a fn in utlis or something and just call it 🙂

psoleckimoj (dismissed):
LGTM! Although could still move that code into a function 🙂

psoleckimoj (approved):
LGTM!

PR #417, opened 2022-11-16
VB-1388: Enable establishment switcher in staging env

tpmcgowan:
Turn on establishment switcher for staging so that environment can be used for testing.


psoleckimoj (approved):
LGTM!

PR #419, opened 2022-11-17
VB-1401: Remove prisonId from GET /visits call

tpmcgowan:
This removes the `prisonId` parameter when calling `GET /visits` now that this is no longer required by the Visit Scheduler. 

This means that , for example, on a prisoner profile visit history tab visits from any prison will show instead of just those from the selected establishment.


psoleckimoj (approved):
LGTM!

PR #420, opened 2022-11-21
VB-1386, add prison specific content to select visitors

hutcheonb-moj:
## Description
Add prison specific visitor number information to the select visitors page.
Previously decided this should be hard coded in the front end, instead of being served by the back end.

### DIFF
--- a/server/routes/visitJourney/selectVisitors.test.ts
+++ b/server/routes/visitJourney/selectVisitors.test.ts
@@ -422,6 +422,58 @@ testJourneys.forEach(journey => {
           expect($('[data-test="back-to-start"]').length).toBe(1)
         })
     })
+
+    it('should display prison specific content, related to Hewell prison', () => {
+      sessionApp = appWithAllRoutes({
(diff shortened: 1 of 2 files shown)


server/routes/visitJourney/selectVisitors.test.ts
```diff
@@ -422,6 +422,47 @@ testJourneys.forEach(journey => {
           expect($('[data-test="back-to-start"]').length).toBe(1)
         })
     })
+
+    describe(`Display prison specific content for each prison`, () => {
+      const prisons = [
+        {
+          prisonId: 'HEI',
+          prisonName: 'Hewell (HMP)',
+          content1: 'Add up to 3 people aged 10 and over, and 4 children under 10 years old.',
```
psoleckimoj, line 431: Not sure where `contentX` came from but would be nice if it was descriptive about where that text sits?

psoleckimoj (approved):
Other than my comment which can be sorted after, LGTM!

psoleckimoj (approved):
LGTM

PR #422, opened 2022-11-21
VB-1409: Use Prison Register API to get list of all prisons (with caching)

tpmcgowan:
This change adds support for the [Prison Register API](https://prison-register-dev.hmpps.service.justice.gov.uk/swagger-ui/index.html) to get a list of all prisons, which is used to find prison names for supported prisons.

The list of all prisons is cached within the `SupportedPrisonsService` for 24 hours to avoid repeated calls to the API to retrieve rarely changing data. 

(Following on from this I will be changing the data format used to pass Prison data within the application from an array of `Prison` objects to a single object with key/values of `prisonId / prisonName` to simplify code and have less repeated searching through arrays. Keeping separate to make PRs manageable.)


server/services/supportedPrisonsService.ts
```diff
@@ -1,25 +1,34 @@
 import { Prison, SystemToken } from '../@types/bapv'
-import prisons from '../constants/prisons'
+import PrisonRegisterApiClient from '../data/prisonRegisterApiClient'
 import VisitSchedulerApiClient from '../data/visitSchedulerApiClient'
 
 type VisitSchedulerApiClientBuilder = (token: string) => VisitSchedulerApiClient
+type PrisonRegisterApiClientBuilder = (token: string) => PrisonRegisterApiClient
+
+const A_DAY_IN_MS = 24 * 60 * 60 * 1000
```
tpmcgowan, line 8: I've set it to cache for a day as I'm assuming it barely ever changes, but open to suggestions for shorter/longer time 🤷‍♂️ !

psoleckimoj (approved):
LGTM!

PR #423, opened 2022-11-22
VB-1409: Use supported prisons list from Visit Scheduler

tpmcgowan:
Use list of supported prison IDs returned from the Visit Scheduler.


psoleckimoj (approved):
LGTM!

PR #424, opened 2022-11-22
VB-1450 audit event - change establishment

hutcheonb-moj:
## Description
Added audit event messages to successful establishment changes.
Logs the old and new establishment.

The second part of the ticket has already been covered, the only audit message that doesn't use currentEstablishment is visit search, which should show the visits prison, not the users.


tpmcgowan (approved):
LGTM!

PR #426, opened 2022-11-25
VB-1449, load session data when updating visit, not viewing

hutcheonb-moj:
Session data is now populated when entering the update journey, instead of when viewing a visit.

test data is used for both GET and POST methods, instead of having copies in each describe block


psoleckimoj (approved):
LGTM!

PR #427, opened 2022-11-28
VB-1412: Default selected establishment to user's active caseload

tpmcgowan:
This change:
* uses the user's caseload (set in DPS) to set the default selected establishment
* redirects the user to choose an establishment if the active caseload is not a supported VSiP prison
* maintains the already-selected prison in VSiP if the caseload in DPS is changed

If `FEATURE_ESTABLISHMENT_SWITCHER_ENABLED` is not enabled then the establishment defaults to `HEI / Hewell` regardless of the user's active caseload.

### DIFF
--- a/server/middleware/populateSelectedEstablishment.ts
+++ b/server/middleware/populateSelectedEstablishment.ts
@@ -1,14 +1,30 @@
-import type { NextFunction, Request, Response } from 'express'
-import { Prison } from '../@types/bapv'
+import type { RequestHandler } from 'express'
+import config from '../config'
+import SupportedPrisonsService from '../services/supportedPrisonsService'
 
-// temporarily hard-coding here pending work on establishment switcher and decision on default value
-const defaultEstablishment: Prison = { prisonId: 'HEI', prisonName: 'Hewell (HMP)' }
+export default function populateSelectedEstablishment(
+  supportedPrisonsService: SupportedPrisonsService,
+): RequestHandler {
+  return async (req, res, next) => {
+    // using req.originalUrl rather than ideally req.path as this was causing problems
+    // because of middleware sometimes being called twice (expected to be resolved in VB-1430)
+    if (req.session.selectedEstablishment === undefined && !req.originalUrl.startsWith('/change-establishment')) {
+      const supportedPrisons = await supportedPrisonsService.getSupportedPrisons(res.locals.user?.username)
 
-export default function populateSelectedEstablishment(req: Request, res: Response, next: NextFunction) {
-  if (req.session.selectedEstablishment === undefined) {
-    req.session.selectedEstablishment = defaultEstablishment
+      // Override active caseload with Hewell if establishment switcher feature not enabled
+      const activeCaseLoadId = config.features.establishmentSwitcherEnabled ? res.locals.user.activeCaseLoadId : 'HEI'
+
+      if (supportedPrisons[activeCaseLoadId]) {
+        req.session.selectedEstablishment = {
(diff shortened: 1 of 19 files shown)


server/routes/testutils/appSetup.ts
```diff
@@ -53,6 +56,16 @@ class MockUserService extends UserService {
   }
 }
 
+class MockSupportedPrisonsService extends SupportedPrisonsService {
```
tpmcgowan, line 59: Using this mock service here only for passing to the `standardRouter` so that the `populatedSelectedEstablishment` middleware can work for `supertest` tests on routes.

Will have to be revisited when looking at refactoring along the lines of https://github.com/ministryofjustice/hmpps-template-typescript/pull/106.

server/routes/testutils/appSetup.ts
```diff
@@ -94,7 +107,7 @@ function appSetup({
 
   app.use((req, res, next) => {
     res.locals = {}
-    res.locals.user = req.user
+    res.locals.user = user
```
tpmcgowan, line 110: Bug here that was fixed in the template app soon after we started (https://github.com/ministryofjustice/hmpps-template-typescript/pull/62)! The `user` defined above in this file for testing wasn't actually being used.

This change has led to lots of other test files in this PR having changes to 'user' from `undefined` to `user1`.

server/services/auditService.ts
```diff
@@ -347,7 +347,7 @@ export default class AuditService {
       operationId,
       who,
       service: config.apis.audit.serviceName,
-      details: JSON.stringify(details),
+      details: JSON.stringify(details, this.replaceUndefinedWithNull),
```
tpmcgowan, line 350: `undefined` is not valid in JSON so such values were 'disappearing' from user audit messages. E.g. change establishment log when first logging in was losing `previousEstablishment: undefined`.

This fix replaces any `undefined` values that might crop up in audit messages with `null` so they do get logged.

server/middleware/populateSelectedEstablishment.ts
```diff
@@ -1,14 +1,31 @@
...
-export default function populateSelectedEstablishment(req: Request, res: Response, next: NextFunction) {
-  if (req.session.selectedEstablishment === undefined) {
-    req.session.selectedEstablishment = defaultEstablishment
-  }
-  res.locals.selectedEstablishment = req.session.selectedEstablishment
+      // Override active caseload with Hewell if establishment switcher feature not enabled
+      const activeCaseLoadId = config.features.establishmentSwitcherEnabled ? res.locals.user.activeCaseLoadId : 'HEI'
 
-  next()
+      if (supportedPrisons[activeCaseLoadId]) {
```
psoleckimoj, line 18: I'm being picky but could you not reverse the if/else, that way you wouldn't need the else as the `if` would be doing a `return` so no need for the else?

server/services/auditService.ts
psoleckimoj: Nice 👍

server/routes/testutils/appSetup.ts
psoleckimoj: Well spotted!

psoleckimoj (approved):
LGTM!

server/middleware/populateSelectedEstablishment.ts
tpmcgowan: Good suggestion! Have just pushed up a change with this in.

psoleckimoj (approved):
LGTM!

PR #429, opened 2022-11-30
VB-1481: Store prison ID in visit session data and check against selected establishment

tpmcgowan:
The existing `visitSessionData` didn't store the `prisonId` associated with the visit and instead relied on the `selectedEstablishment`. When looking up a visit by reference and with a different establishment selected to that of the visit, it was possible to start an update journey for the visit in the context of the 'wrong' prison (VB-1480).

This change stores the `prisonId` in `visitSessionData` at the start of the booking / update journey. This is then checked against the `selectedEstablishment` wherever `sessionCheckMiddleware` is used to make sure the two don't become mismatched. If they do, the request is redirected back to the home page.

Further work is expected to avoid this kind of situation occurring and/or handling it better by changing how the look up visit by reference works; this PR is intended just to be 'fail-safe' against changing a visit for the 'wrong' prison in any scenario.

(Most of the changes in this PR are just adding the new, required `prisonId` to test data. Main change is in the `sessionCheckMiddleWare`.)

### DIFF
--- a/server/@types/bapv.d.ts
+++ b/server/@types/bapv.d.ts
@@ -188,6 +188,7 @@ export type VisitSessionData = {
     restrictions?: OffenderRestriction[]
     previousRestrictions?: OffenderRestriction[]
   }
+  prisonId: string
   visitSlot?: VisitSlot
   originalVisitSlot?: VisitSlot
   visitRestriction?: Visit['visitRestriction']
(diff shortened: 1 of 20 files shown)


tpmcgowan:
Closing in favour of adding `prisonId` to `VisitSlot` instead (#430).

PR #431, opened 2022-12-01
VB-1481: Restrict visit reference search to selected establishment

tpmcgowan:
This change prevents the 'Search for a booking by reference' page from displaying a visit if is for a prison that is not the currently selected establishment.


psoleckimoj (approved):
LGTM!

PR #432, opened 2022-12-02
VB-1481: Add selected establishment check to booking summary page

tpmcgowan:
On the booking summary page, do not show visit details if the visit is for a different prison to the user's currently selected establishment. Instead, show message saying what prison the visit is booked for and prompting user to switch establishment and search for it.

Applies the same check on POSTing to this page (i.e. starting a visit update journey).

(Also includes minor tweak to cancellation SMS to look up the prison name from the visit's `prisonId` rather than `selectedEstablishment` to avoid any potential mix-up.)


hutcheonb-moj (approved):
LGTM

PR #434, opened 2022-12-06
VB-1421: Set case load for DPS when changing establishment

tpmcgowan:
When changing establishment, also update the user's active case load so that this is reflected in, for example, DPS. Achieves this by calling `/api/users/me/activeCaseLoad` in the [Prison API](https://api-dev.prison.service.justice.gov.uk/swagger-ui/index.html#/users/updateMyActiveCaseLoad).

Caveats:
* if you are already logged into DPS and change establishment within VSiP, the selected establishment won't change in DPS until you next log in because the currently active case load is only looked up on login
* likewise, once logged into VSiP, any change of case load in DPS won't affect VSiP until next login

### DIFF
--- a/server/routes/changeEstablishment.test.ts
+++ b/server/routes/changeEstablishment.test.ts
@@ -174,6 +174,7 @@ describe('POST /change-establishment', () => {
           username: 'user1',
           operationId: undefined,
         })
+        // @TODO should also check setActiveCaseLoad is called (awaiting VB-1430)
       })
   })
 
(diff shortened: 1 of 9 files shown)


server/routes/changeEstablishment.test.ts
```diff
@@ -174,6 +174,7 @@ describe('POST /change-establishment', () => {
           username: 'user1',
           operationId: undefined,
         })
+        // @TODO should also check setActiveCaseLoad is called (awaiting VB-1430)
```
tpmcgowan, line 177: Ideally would check case load gets set here but got bogged down trying to do so because of the way `MockUserService` is used in `appSetup.ts` - figured this would be better sorted out once work to refactor to use a services container is done (VB-1430)

hutcheonb-moj (approved):
LGTM

PR #439, opened 2022-12-08
VB-1497: Visits by date - temp fix for capacity counts

tpmcgowan:
'View visits by date' page shows capacity counts hard-coded for Hewell. This change caters for Bristol having different values.

This is a temporary workaround until functionality added to calculate the capacity from the session template for the day.

### DIFF
--- a/server/routes/visits.test.ts
+++ b/server/routes/visits.test.ts
@@ -2,12 +2,14 @@ import type { Express } from 'express'
 import request from 'supertest'
 import * as cheerio from 'cheerio'
 import { format } from 'date-fns'
+import { SessionData } from 'express-session'
 import PrisonerSearchService from '../services/prisonerSearchService'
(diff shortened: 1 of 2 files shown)


psoleckimoj (approved):
LGTM!

rj-adams (commented):
I would prefer visits.ts to explicitly state both Hewell and Bristol - and have something else as a default - using low non-livelike numbers.

This will mean that we can still test other prisons but won't have a false sense of security that all is ok when we actually introduce further prisons - we will be reminded to do something more permanent

rj-adams:
ideally now, we should have no Hewell-esque defaults anywhere in the code

tpmcgowan:
Sure, I can do that (as a new PR as this one's been merged). I'll put the default as `0` as that should hopefully be an obvious prompt!

Won't be there for long in any case as I'll look to calculate this properly as we discussed earlier as part of VB-1497.

PR #440, opened 2022-12-08
VB-1497: Explicitly set Hewell, Bristol and default capacity values

tpmcgowan:
Explicitly set capacity counts for Hewell, Bristol and 'default' on the 'View visits by date' page (change suggested as feedback #439)

(Still a temporary fix while working on how to derive these from session templates.)

### DIFF
--- a/server/routes/visits.ts
+++ b/server/routes/visits.ts
@@ -56,8 +56,14 @@ export default function routes(
     }
 
     // VB-1497 - temporary workaround for capacity counts for Hewell / Bristol
-    const maxSlotDefaults =
-      prisonId === 'HEI' ? { OPEN: 30, CLOSED: 3, UNKNOWN: 30 } : { OPEN: 20, CLOSED: 1, UNKNOWN: 20 }
+    let maxSlotDefaults: { OPEN: number; CLOSED: number; UNKNOWN: number }
+    if (prisonId === 'HEI') {
+      maxSlotDefaults = { OPEN: 30, CLOSED: 3, UNKNOWN: 30 }
+    } else if (prisonId === 'BLI') {
+      maxSlotDefaults = { OPEN: 20, CLOSED: 1, UNKNOWN: 20 }
+    } else {
+      maxSlotDefaults = { OPEN: 0, CLOSED: 0, UNKNOWN: 0 }
+    }
 
     const maxSlots = maxSlotDefaults[visitType] ?? 0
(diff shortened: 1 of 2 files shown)


server/routes/visits.ts
```diff
@@ -56,8 +56,14 @@ export default function routes(
...
     // VB-1497 - temporary workaround for capacity counts for Hewell / Bristol
-    const maxSlotDefaults =
-      prisonId === 'HEI' ? { OPEN: 30, CLOSED: 3, UNKNOWN: 30 } : { OPEN: 20, CLOSED: 1, UNKNOWN: 20 }
+    let maxSlotDefaults: { OPEN: number; CLOSED: number; UNKNOWN: number }
+    if (prisonId === 'HEI') {
+      maxSlotDefaults = { OPEN: 30, CLOSED: 3, UNKNOWN: 30 }
+    } else if (prisonId === 'BLI') {
+      maxSlotDefaults = { OPEN: 20, CLOSED: 1, UNKNOWN: 20 }
+    } else {
+      maxSlotDefaults = { OPEN: 0, CLOSED: 0, UNKNOWN: 0 }
```
psoleckimoj, line 65: Could actually put defaults in the initial `let` to save the `else`? Will approve but would make it tidier 🙂

psoleckimoj (dismissed):
LGTM!

server/routes/visits.ts
rj-adams: Looks fine - but why do we need a value for UNKNOWN here?

tpmcgowan: > Looks fine - but why do we need a value for UNKNOWN here?

Not sure - haven't looked into this. I think @psoleckimoj added this to handle case where viewing visits in the past might have this visit restriction value?

psoleckimoj: Yep, it all falls over without Unknown as all past visits are unknown.

psoleckimoj (approved):
LGTM!

server/routes/visits.ts
tpmcgowan: 👍 - looks like https://dsdmoj.atlassian.net/browse/VB-1162 has the details

rj-adams: sorry - any chance we could make the defaults 2,1,0 or similar - i.e. not live-like but we are able to introduce new prisons for testing etc without having to change code

PR #443, opened 2022-12-15
VB-1462, Remove establishment switcher feature flag

hutcheonb-moj:
## Description
Remove establishment switcher feature flag checks.
Also removed the variable from config and helm files


tpmcgowan (changes requested):
Looks great! Just spotted one minor thing: `feature.env` - which is the `.env` file used for integration (Cypress) tests - has a stray `FEATURE_ESTABLISHMENT_SWITCHER_ENABLED: "true"`

tpmcgowan (approved):
LGTM!

PR #447, opened 2022-12-21
VB-1540, remove supportTypes duplication, added to testObject

hutcheonb-moj:
createSupportTypes() can now be called in place of all availableSupportType declarations. 
No generator feature was used as these are always currently just carbon copy definitions of available support types


tpmcgowan (approved):
LGTM!

PR #448, opened 2022-12-21
VB-1540, moved prisoner to test data - prisoner profile tests

hutcheonb-moj:
Removed the basic prisoner from the prisoner profile tests, added it into the testObjects file.
This test file still needs more work

### DIFF
--- a/server/data/__testutils/testObjects.ts
+++ b/server/data/__testutils/testObjects.ts
@@ -1,6 +1,6 @@
 import { Prison } from '../../@types/bapv'
-import { CaseLoad } from '../prisonApiTypes'
 import { SupportType } from '../visitSchedulerApiTypes'
+import { InmateDetail, CaseLoad } from '../prisonApiTypes'
 
 export const createCaseLoads = ({
   caseLoads = [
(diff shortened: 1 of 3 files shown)


server/data/__testutils/testObjects.ts
```diff
@@ -67,3 +67,22 @@ export const createSupportTypes = ({
     },
   ] as SupportType[],
 } = {}): SupportType[] => supportTypes
+
+export const createPrisonerProfile = ({
```
tpmcgowan, line 71: Just wondering about the name and whether `createPrisonerProfile` risks being confused with the specific `PrisonerProfile` type we have ([here](https://github.com/ministryofjustice/book-a-prison-visit-staff-ui/blob/9aded01a1d26dafcd31f7f0bda59f5594ebfe179/server/%40types/bapv.d.ts#L122)) that's used to pass all the data for the Prisoner Profile page.

Perhaps something like just `createPrisoner`, `createInmate` or just be explicit with `createInmateDetail`? Tricky I know because we've a lot of overlapping terms!

tpmcgowan (commented):
Looks good 🙂 - just a minor comment on naming to have a think about

tpmcgowan (approved):
LGTM!

PR #449, opened 2022-12-22
VB-1547: Trim whitespace - visit cancellation reason & main contact name

tpmcgowan:
Trim whitespace so it isn't sent back to the visit scheduler when:
* entering a visit cancellation reason
* entering a custom main contact name


AledWEvans (approved):

hutcheonb-moj (approved):
LGTM

PR #452, opened 2022-12-30
VB-1565, add Bristol+Hewell Prison phone numbers to cancellation

hutcheonb-moj:
## Description
Cancellation SMS had test data being sent out as the prison phone number.
Updated the Cancellation SMS function call to include the correct prison phone number, based on which prison the visit was booked for

### DIFF
--- a/server/routes/visit.test.ts
+++ b/server/routes/visit.test.ts
@@ -759,7 +759,30 @@ describe('POST /visit/:reference/cancel', () => {
           phoneNumber: cancelledVisit.visitContact.telephone,
           visitSlot: cancelledVisit.startTimestamp,
           prisonName: 'Hewell (HMP)',
-          prisonPhoneNumber: '01234443225',
+          prisonPhoneNumber: '0300 060 6503',
+        })
+      })
+  })
+
+  it('should send the SMS with the correct prison phone number - Bristol', () => {
+    cancelledVisit.prisonId = 'BLI'
+    config.apis.notifications.enabled = true
+
+    return request(app)
+      .post('/visit/ab-cd-ef-gh/cancel')
+      .send('cancel=PRISONER_CANCELLED')
+      .send('reason_prisoner_cancelled=illness')
+      .expect(302)
+      .expect('location', '/visit/cancelled')
+      .expect(() => {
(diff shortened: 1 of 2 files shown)


server/routes/visit.test.ts
```diff
@@ -759,7 +759,30 @@ describe('POST /visit/:reference/cancel', () => {
...
+      .expect('location', '/visit/cancelled')
+      .expect(() => {
+        expect(visitSessionsService.cancelVisit).toHaveBeenCalledTimes(1)
+        expect(auditService.cancelledVisit).toHaveBeenCalledTimes(1)
+        expect(notificationsService.sendCancellationSms).toHaveBeenCalledTimes(1)
+        expect(notificationsService.sendCancellationSms).toHaveBeenCalledWith({
+          phoneNumber: cancelledVisit.visitContact.telephone,
+          visitSlot: cancelledVisit.startTimestamp,
+          prisonName: 'Bristol (HMP & YOI)',
+          prisonPhoneNumber: '0300 060 6510',
```
psoleckimoj, line 785: Only suggestions is having these in a config file or somewhere like that. I realise data probably won't change but makes it easier to edit tests and the config can be used in the test and the code. Finally once the data is in a config file, you can probably iterate round it in the tests to save copying the same test for each site.

psoleckimoj (approved):
LGTM but needs another PR to look at what I suggested

server/routes/visit.ts
```diff
@@ -323,11 +323,19 @@ export default function routes(
           const supportedPrisons = await supportedPrisonsService.getSupportedPrisons(res.locals.user?.username)
           const prisonName = supportedPrisons[visit.prisonId]
 
+          // Current prison phone number configuration, look to move in the future
+          let prisonPhoneNumber = ''
+          if (prisonName === 'Hewell (HMP)') {
```
tpmcgowan, line 328: Probably better to test against `visit.prisonId` (e.g. `HEI`) - but that can be addressed when putting stuff in a config file as Paul suggested (filed as [VB-1594](https://dsdmoj.atlassian.net/browse/VB-1594))

tpmcgowan (approved):
LGTM

PR #453, opened 2023-01-03
Update dependencies 2023-01-03

psoleckimoj:


hutcheonb-moj:
Ive updated the way the tests are done for displayAge.
The displayAge method requires more work, but I've included all the tests that need to be run to check all the possible cases and problems where the code will fail (Commented out the currently failing test data)
Also changed the DOB in the visit.test file, this will need a more permanent solution that I'll create a ticket for, but in the mean time this gets the tests passing without any major changes, in order to get these dependencies done

hutcheonb-moj (approved):
LGTM, will create a ticket to go through the visit.test and displayAge method

tpmcgowan (approved):
LGTM - with the caveat that @hutcheonb-moj is creating a ticket to properly fix the `displayAge` filter and tests

PR #456, opened 2023-01-04
VB-1600, Add reference to cancellation SMS

hutcheonb-moj:
## Description
Add visit reference to Cancellation SMS to be used in Gov Notify.
Also changes the naming of a test data in notificationsApiClient.test as it was misleading

### DIFF
--- a/server/data/notificationsApiClient.ts
+++ b/server/data/notificationsApiClient.ts
@@ -51,12 +51,14 @@ class NotificationsApiClient {
     visitTime,
     visitDate,
     prisonPhoneNumber,
+    reference,
   }: {
     phoneNumber: string
     prisonName: string
     visitTime: string
     visitDate: string
     prisonPhoneNumber: string
+    reference: string
   }): Promise<SmsResponse> {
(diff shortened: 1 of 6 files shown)


server/data/notificationsApiClient.ts
```diff
@@ -51,19 +51,22 @@ class NotificationsApiClient {
...
     prisonPhoneNumber: string
+    reference: string
   }): Promise<SmsResponse> {
     return this.notificationsApiClient.sendSms(templates.cancellationConfirmation, phoneNumber, {
       personalisation: {
         prison: prisonName,
         time: visitTime,
         date: visitDate,
         'prison phone number': prisonPhoneNumber,
+        reference,
```
tpmcgowan, line 69: This means it will need to be referenced in the Notify as `((reference))` rather than `((ref number))` like the others. Doesn't particularly matter - just need to be aware and use the right one. Personally I think it's better not to use spaces in these 🤷‍♂️

tpmcgowan (approved):
LGTM!

PR #457, opened 2023-01-09
VB-1593, manual age calculation

hutcheonb-moj:
## Description
DisplayAge's calculation was found to have been giving incorrect ages in certain situations.
Attempted to use Date-FNS (which is already being used in this project), but this also failed some specific situations.

This is a manual method for calculating ages, which works with both Plural and Singular.
It's fiddly and not the easiest to read through, but it works on all tests (some of which were previously found to cause issues with previous iterations of this logic)

Very open to suggestions as this is my own calculation (Not copy pasted)

### DIFF
--- a/server/utils/nunjucksSetup.ts
+++ b/server/utils/nunjucksSetup.ts
@@ -85,23 +85,30 @@ export function registerNunjucks(app?: express.Express): Environment {
   })
 
   njkEnv.addFilter('displayAge', (dateOfBirth: string) => {
-    let ageString
-
     const today = new Date()
     const dob = new Date(dateOfBirth)
-    const years = today.getFullYear() - dob.getFullYear()
-    if (years > 1) {
+
+    let ageString = ''
+
+    let years = today.getFullYear() - dob.getFullYear()
+    let months = years * 12 + (today.getMonth() - dob.getMonth())
+    const days = today.getDate() - dob.getDate()
+
+    if (days < 0) {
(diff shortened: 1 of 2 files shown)


server/utils/nunjucksSetup.ts
```diff
@@ -85,23 +85,30 @@ export function registerNunjucks(app?: express.Express): Environment {
   })
 
   njkEnv.addFilter('displayAge', (dateOfBirth: string) => {
-    let ageString
-
     const today = new Date()
     const dob = new Date(dateOfBirth)
-    const years = today.getFullYear() - dob.getFullYear()
-    if (years > 1) {
+
```
psoleckimoj, line 90: It just occurred to me, is there a reason you're not using the built-in `date-fn` library we use on the project. I really have to apologise for not spotting that!!!

psoleckimoj: https://date-fns.org/docs/Getting-Started

hutcheonb-moj: mentioned in the description. Tried it with date fns but that also leads you to have to make these work-arounds with particular dates (It struggled with months) 
That's what led me to make this manual calculation instead of part Date FNS, part our own

psoleckimoj: Fair enough!

psoleckimoj (approved):
LGTM!

PR #459, opened 2023-01-10
VB-1593: Implement 'displayAge' with date-fns

tpmcgowan:


server/utils/nunjucksSetup.test.ts
```diff
@@ -124,14 +124,14 @@ describe('Nunjucks Filters', () => {
 
   describe('displayAge', () => {
     beforeAll(() => {
-      const fakeDate = new Date('2020-12-14')
+      const fakeDate = new Date('2020-12-14T12:00:00')
```
tpmcgowan, line 127: I found setting the time to actually during the day fixed some of the incorrect comparisons - I think perhaps otherwise it defaults to midnight and it doesn't give the right answer on that boundary

hutcheonb-moj (commented):
LGTM, tried adding some random dates and appears to cover everything.
Would recommend taking this through to production in place of my implementation

tpmcgowan:
OK, thanks for testing, @hutcheonb-moj . Have marked as ready for review

hutcheonb-moj (approved):
LGTM

PR #460, opened 2023-01-11
VB-1463, View details of bookings - prisoner profile page

hutcheonb-moj:
## Description
Update prisoner profile page to include:
Reference with link to a specific booking
Both booked and cancelled visits are to be shown


tpmcgowan (approved):
LGTM!

PR #461, opened 2023-01-11
Add script to update API type defs

tpmcgowan:
* Add a script (`./bin/update-types.sh`) that uses `openapi-typescript` to download and update TypeScript files for APIs used in the project. It also tidies and formats them to pass linting rules.
* Document this in the README
* Update existing type definitions (and renamed some for consistency)


psoleckimoj (approved):
LGTM! and nice idea 👍

PR #463, opened 2023-01-12
VB-1551: Use Prisoner Offender Search for incentive level

tpmcgowan:
* Switch to using Prisoner Offender Search to retrieve incentive level (replacing Prison API)
* Add more methods for creating standard test data objects and refactored existing tests to use these - most changes in the PR actually relate to this


integration_tests/e2e/searchForAPrisoner.cy.ts
```diff
@@ -188,6 +188,7 @@ context('Search for a prisoner', () => {
         .then(pageTitle => {
           cy.task('stubGetBookings', prisonerNumber)
           cy.task('stubGetOffender', offender)
+          cy.task('stubGetPrisonerById', { prisonerNumber, currentIncentive: { level: { description: 'Standard' } } })
```
tpmcgowan, line 191: Intending to revise how these stubs and the mock data works for the integration tests as part of VB-729.

server/data/__testutils/testObjects.ts
```diff
@@ -21,6 +22,83 @@ export const createCaseLoads = ({
...
 } = {}): CaseLoad[] => caseLoads
 
+export const createCurrentIncentive = ({
+  level = {
+    code: 'STD',
+    description: 'Standard',
+  },
+}: Partial<CurrentIncentive> = {}): CurrentIncentive => ({ level } as CurrentIncentive)
+
+export const createInmateDetail = ({
```
tpmcgowan, line 32: Essentially, we have here the three different kinds of representations we have for the same prisoner, coming from different API endpoints.

hutcheonb-moj (approved):
ltgm

server/data/__testutils/testObjects.ts
psoleckimoj: @tpmcgowan could use this for the moving off NOMIS team about standardising data 😄

server/data/prisonerSearchClient.test.ts
```diff
@@ -79,21 +79,16 @@ describe('prisonSearchClientBuilder', () => {
...
-        prisonId: 'HEI',
-        prisonName: 'HMP Hewell',
-        cellLocation: '1-1-C-028',
-        restrictedPatient: false,
-      } as Prisoner
-      fakePrisonerSearchApi.get('/prisoner/A1234BC').matchHeader('authorization', `Bearer ${token}`).reply(200, results)
+      const prisoner = createPrisoner()
+
+      fakePrisonerSearchApi
+        .get('/prisoner/A1234BC')
```
psoleckimoj, line 85: My only suggestion here would be there is no link between the prisonerId in the get and createPrisoner, maybe if you passed an id to the latter it creates more of an understandable link? If that makes sense?

server/services/prisonerProfileService.test.ts
```diff
@@ -449,21 +450,19 @@ describe('Prisoner profile service', () => {
...
+      const inmateDetail = createInmateDetail({
         activeAlertCount: 4,
         inactiveAlertCount: 1,
-        legalStatus: 'REMAND',
         alerts: [inactiveAlert, nonRelevantAlert, ...alertsToFlag],
-      }
+        legalStatus: 'REMAND',
+      })
+
+      const prisoner = createPrisoner()
```
psoleckimoj, line 461: Or maybe use the prisoner object as the expected results parts? Otherwise if you change the DoB in the standard prisoner you have to change it every test that checks it, if you say it should equal `prisoner.dob` or similar you don't need to do that? Or some similar way of doing that?

server/data/prisonerSearchClient.test.ts
tpmcgowan: Yeah, good suggestion; that would make it clearer. Or, alternatively use something like `.get('/prisoner/${prisoner.offenderNo}')`

Trying to chip away at rationalising and tidying the tests so will add to the list!

psoleckimoj: Aye tis awesome work 👍

server/data/__testutils/testObjects.ts
tpmcgowan: yeah, the data we need does unfortunately seem to fall between several endpoints that all partially overlap 😵‍💫! I've suggested to Al/Dhiraj that something the orchestration service (or similar) could do is return us a 'VSiP Prisoner' - i.e. cobble together what we need from various sources so we just have one representation of that data in the front-end

PR #469, opened 2023-01-16
VB-1458, error message in production - 404

hutcheonb-moj:
## Description
Changed the 404 message being displayed in production.
Changed the way this page was rendering the message, for ease of setting particular message for other error codes in the future. 

Error stack still displays when running locally.


tpmcgowan:
Just noting, @psoleckimoj, in case you take a look -  I've just had a chat about this with @hutcheonb-moj and suggested tweaking it so behaviour (apart from rendering stack trace) is more consistent between `development` and `production`.

server/views/pages/error.njk
```diff
@@ -9,11 +10,25 @@
...
-      <h1 class="govuk-heading-l">{{ pageHeaderTitle }}</h1>
-      <h2 class="govuk-heading-m">{{ status }}</h2>
-      <p>{{ message }}</p>
+      <h1 class="govuk-heading-l">{{ heading }}</h1>
+      {% for item in message %}
+        <p>{{ item }}</p>
+      {% endfor %}
+
+      {% if showHomeButton %}
+        {{ govukButton({
```
tpmcgowan, line 19: Could replace actually rendering the button here with a component we already have:
```
{% include "partials/backToStartButton.njk" %}
```

This has the advantage of the URL being `/back-to-start` which is a route that clears session and lands you on the home page. If making this change, would need to tweak the test as it will perhaps have different data selector and URL

hutcheonb-moj: completely forgot about this partial/route, thanks!

server/errorHandler.ts
```diff
@@ -11,14 +11,21 @@ export default function createErrorHandler(production: boolean) {
       return res.redirect('/sign-out')
     }
 
+    const heading = error.status === 404 ? 'Page not found' : 'Error'
     const prodMessage =
```
tpmcgowan, line 15: Happy to approve, but I guess if being picky on naming this would probably now be better named just `message` to avoid confusion

tpmcgowan (approved):
LGTM! (just minor comment about naming, that you could consider now or when more error pages added...)

PR #470, opened 2023-01-18
VB-1704: Limit number of selected visitors to maximum 10

tpmcgowan:
Add validation to select visitors page to limit maximum to 10. This is to match the limit imposed by the Visit Scheduler.


hutcheonb-moj (approved):
LGTM

PR #473, opened 2023-01-23
VB-1594: Consolidate prison specific config

tpmcgowan:
* Consolidate prison-specific information (phone number, advisory text for visitors numbers, slot capacities) into a simple lookup function.
* It defaults to empty values if a prison hasn't been configured, meaning that new supported prisons can be tested before specific configuration information is available
* Update code and tests to use this information

This change means that adding appropriate configuration information for new prisons now only requires adding entries to `server/constants/prisonConfiguration.ts` rather than changing custom code in multiple places. (Longer-term this may be reworked as the orchestration service may end up proving some of this information.)

### DIFF
--- a/server/constants/prisonConfiguration.test.ts
+++ b/server/constants/prisonConfiguration.test.ts
@@ -1,0 +1,21 @@
+import getPrisonConfiguration from './prisonConfiguration'
+
+describe('Prison configuration', () => {
+  it('should return correct prison configuration for requested prisonId', () => {
+    const prisonConfiguration = getPrisonConfiguration('HEI')
+
+    expect(prisonConfiguration.prisonPhoneNumber).toBe('0300 060 6503')
+    expect(prisonConfiguration.selectVisitorsText.length).toBe(2)
+    expect(prisonConfiguration.visitCapacity.open).toBe(30)
+    expect(prisonConfiguration.visitCapacity.closed).toBe(3)
+  })
+
+  it('should return default prison configuration when none available for requested prisonId', () => {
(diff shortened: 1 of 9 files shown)


server/constants/prisonConfiguration.test.ts
```diff
@@ -0,0 +1,21 @@
+import getPrisonConfiguration from './prisonConfiguration'
+
+describe('Prison configuration', () => {
+  it('should return correct prison configuration for requested prisonId', () => {
+    const prisonConfiguration = getPrisonConfiguration('HEI')
+
+    expect(prisonConfiguration.prisonPhoneNumber).toBe('0300 060 6503')
```
psoleckimoj, line 7: Do you think we should have a results file for these prison specific tests? Just thinking if a phone number changes, it'd be nice to change in one place that every test it's used?

psoleckimoj (approved):
Much nicer way to do it! Getting to a point where we could do with a prison config app 😁 LGTM!

hutcheonb-moj (approved):
LGTM

server/constants/prisonConfiguration.test.ts
tpmcgowan: Not sure ...🤔 Thought about that when I was doing it, but I figured that for testing the config lookup for a `prisonId` it was enough here to do a known & unknown one. Could do this though when we add more prisons, I'll have a think!

(Scattered around the other tests is coverage for Hewell, Bristol and 'unknown' prison config but, again, wasn't planning to expand this to more prisons as it seemed like enough coverage.)

PR #476, opened 2023-01-24
VB-1731, overlapping sessions bug fix

hutcheonb-moj:
## Description
Error caused by overlapping sessions in Cardiff's visit schedule
Caused by the fact there is a session with no capacity for the currently selected visit restriction (the sessions have either open or closed, not both)
Current iteration only pushes the slot on a boolean check,  testing whether the current restriction has any capacity on the slot (initial capacity not available tables)


tpmcgowan (approved):
LGTM!

psoleckimoj (approved):
LGTM!

PR #479, opened 2023-01-27
VB-1740: Remove hard-coded session capacity

tpmcgowan:
Removing hard-coded session capacity data now that this is retrieved from the visit scheduler.


psoleckimoj (approved):
LGTM!

PR #482, opened 2023-01-30
Refactor test data as a class

tpmcgowan:
Refactor existing test data to use a class so it can be brought into tests as a single import, with everything then available. Also moved to `./server/routes/testutils` so it's alongside the related existing test setup `appSetup.ts`.

A lot of changes in this PR but most are just updating imports!


hutcheonb-moj (approved):
LGTM

server/routes/testutils/testData.ts
```diff
@@ -0,0 +1,216 @@
+import { Prison } from '../../@types/bapv'
+import { SessionCapacity, SupportType, Visit } from '../../data/visitSchedulerApiTypes'
+import { InmateDetail, CaseLoad, PrisonerBookingSummary } from '../../data/prisonApiTypes'
+import { CurrentIncentive, Prisoner } from '../../data/prisonerOffenderSearchTypes'
+
+export default class TestData {
```
psoleckimoj, line 6: Do we want to use `static` methods? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/static

tpmcgowan: Yes!

psoleckimoj (approved):
LGTM!

PR #487, opened 2023-02-02
VB-1751: Allow bookings to proceed if Whereabouts API is unavailable

tpmcgowan:
* catches errors from the whereabouts API and handles these by showing message informing user that the data is unavailable, but booking can proceed
* previous behaviour would cause errors from whereabouts to 'bubble up' and prevent user from continuing


server/services/visitSessionsService.ts
```diff
@@ -120,11 +121,18 @@ export default class VisitSessionsService {
...
+    let prisonerEvents: ScheduledEvent[]
+    let whereaboutsAvailable = true
+    try {
+      prisonerEvents = await whereaboutsApiClient.getEvents(
+        offenderNo,
+        format(earliestStartTime, 'yyyy-MM-dd'),
+        format(latestEndTime, 'yyyy-MM-dd'),
+      )
+    } catch (error) {
+      prisonerEvents = []
```
psoleckimoj, line 133: Is this required as you could default it to empty on line 124?

psoleckimoj (dismissed):
LGTM!

psoleckimoj (approved):
LGTM!

PR #489, opened 2023-02-07
VB-1815, added word-break to all govuk-table cells, to fix long string issue

hutcheonb-moj:
## Description
On active alerts, when a long string with no spaces had been added in comments, the string continued across the page, breaking out of its box and pushing all other content right, introducing side scrolling.

`overflow-wrap: anywhere`  is being used as initial suggestion had been deprecated


assets/sass/local.scss
```diff
@@ -247,3 +247,7 @@
     margin-bottom: govuk-spacing(3);
   }
 }
+
+.govuk-table__cell {
+  word-break: break-word;
+}
```
psoleckimoj, line 253: Looks good, just need a newline leaving at the end of the file 👍

psoleckimoj (dismissed):
LGTM!

assets/sass/local.scss
```diff
@@ -247,3 +247,7 @@
     margin-bottom: govuk-spacing(3);
   }
 }
+
+.govuk-table__cell {
+ overflow-wrap: anywhere;
```
tpmcgowan, line 252: Tried this out locally and it's causing words to break mid-way through where it shouldn't happen (see screenshot). Depends a lot on window size so you don't always see it. Might need to be more targeted to the specific columns/cells likely to contain this kind of data?


Screenshot 2023-02-10 at 08 10 40

tpmcgowan (changes requested):
I think this needs tweaking so it doesn't cause 'real' words to break in other cells (see comment with screenshot)

tpmcgowan (approved):
LGTM!

PR #490, opened 2023-02-09
VB-1416 add session conflict message

hutcheonb-moj:
## Description
Add new message when previous visit slot is no longer available 
(IEP Level change, location change, loss of enhanced)

All previous tests affected have been updates, still needs test for this new message

### DIFF
--- a/server/routes/visitJourney/dateAndTime.test.ts
+++ b/server/routes/visitJourney/dateAndTime.test.ts
@@ -615,4 +615,51 @@ describe('Update journey specific warning messages', () => {
         expect($('input#1').prop('checked')).toBe(true)
       })
   })
+
+  it('display both slot information blocks, when slot is no longer available and restriction type has changed', () => {
+    // Changed slot time, so the pre-booked visit can't be matched with a slot
+    slotsList = {
+      'October 2022': [
+        {
+          date: 'Monday 17 October',
+          prisonerEvents: {
+            morning: [],
+            afternoon: [],
+          },
+          slots: {
+            morning: [
+              {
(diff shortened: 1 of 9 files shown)


server/routes/visitJourney/dateAndTime.test.ts
```diff
@@ -531,59 +508,59 @@ describe('Update journey specific warning messages', () => {
...
+        expect($('input#1').prop('checked')).toBe(true)
       })
   })
 
-  it('should show message with no slot selected when visit has changed from open to closed (prisoner restriction), but original timeslot unavailable', () => {
+  it('should show two messages with no slot selected when visit has changed from open to closed and original time slot unavailable', () => {
     currentlyBookedSlot.visitRestriction = 'OPEN'
 
-    currentlyAvailableSlots[0].availableTables = 0
+    currentlyAvailableSlots[0].startTimestamp = '2022-10-17T09:01:00'
```
tpmcgowan, line 541: @hutcheonb-moj For info, I've tweaked the tests a bit as I think they weren't quite setting up the scenarios correctly. For each one in this set, I'm checking the presence/absence of **both** of the possible messages. I'm also setting up the situation where the original time slot isn't available by changing the time. I'm not using the `availableTables` as this is effectively ignored - as even if the slot is overbooked we let an update proceed because it's _already_ in the count of bookings. 

Initially I thought the tests were right but there was something wrong with the implementation - I then realised that the implementation worked but the tests weren't quite doing what the ought to :swear

tpmcgowan (approved):
LGTM

PR #491, opened 2023-02-13
VB-1830: Use visits orchestration service in dev environment

tpmcgowan:
Use visits orchestration service instead of visit scheduler for dev environment only. To enable testing that new service is successfully providing all endpoints.

(Pre-cursor to moving all environments over and re-factoring front-end code to be more appropriately named.)


dhirajshettymoj (approved):
LGTM

PR #492, opened 2023-02-16
VB-1850, view timetable tile

hutcheonb-moj:
## Description 
Added the 'View Timetable' tile to the home page
Added the config to all environments, only enabled in dev.
Updated the integration tests to include testing for the tile.


tpmcgowan (approved):
LGTM (@hutcheonb-moj - I tweaked the tile text as it didn't quite match the ticket.)

PR #497, opened 2023-02-17
VB-1857: Add integration test for user role check

tpmcgowan:
Add integration test to cover user without `ROLE_MANAGE_PRISON_VISITS` role being redirected to auth error page.

(Will add something to `server/routes/testutils/appSetup.ts` to cover roles in unit tests once that's refactored as part of VB-1430.)


psoleckimoj (approved):
LGTM!

PR #498, opened 2023-02-20
VB-1841, Fetch and display visit schedule data

hutcheonb-moj:
## Description
Still draft - on the new timetable page, given a date, the new `/visit-session/schedule` end point is called, this data is then displayed in a table.

### DIFF
--- a/server/views/pages/timetable.njk
+++ b/server/views/pages/timetable.njk
@@ -15,7 +15,6 @@
 
         <h2 class="govuk-heading-m" data-test="date-heading">Monday DD MMMM 2023</h2>
 
-
         <h2 class="govuk-heading-m" data-test="date-heading">Request changes to the timetable</h2>
         <p>You need to provide six weeks' notice to change the timetable. To request a change,
             <a href="LINK_TBC" target="_blank" data-test="change-request">(LINK TBC!) complete the request form (opens in a new tab)</a>.
--- a/server/app.ts
+++ b/server/app.ts
@@ -170,7 +170,18 @@ export default function createApp(userService: UserService): express.Application
     ),
   )
 
-  app.use('/timetable/', timetableRoutes(standardRouter(userService, supportedPrisonsService)))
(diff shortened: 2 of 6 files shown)


server/@types/bapv.d.ts
```diff
@@ -248,3 +248,14 @@ export type VisitsPageSlot = {
   adults: number
   children: number
 }
+
+export type Schedule = {
```
tpmcgowan, line 252: I don't think this is needed? Duplicate of what is happening in `visitSchedulerApiTypes.ts` where it has:
```
export type SessionSchedule = components['schemas']['SessionScheduleDto']
```

hutcheonb-moj: this is a relic from when I was doing the processing at the route level, will get this out tomorrow.

server/routes/timetable.test.ts
```diff
@@ -3,11 +3,24 @@ import request from 'supertest'
...
+const visitSessionsService = new VisitSessionsService(
+  null,
+  null,
+  null,
+  systemToken,
+) as jest.Mocked<VisitSessionsService>
 
 beforeEach(() => {
-  app = appWithAllRoutes({})
+  visitSessionsService.getSessionSchedule.mockResolvedValue([])
```
tpmcgowan, line 21: Tests were failing because this service needed mocking. Just returning empty schedule for now.

server/views/pages/timetable.njk
```diff
@@ -40,11 +74,45 @@
...
+            },
+            {
+              text: "End date"
+            }
+          ],
+          rows: scheduleRows
+        }) }}
+
+        {% else %}
+          <p data-test="empty-schedule">No visit sessions on this day.</p>
```
tpmcgowan, line 112: Added condition to catch empty schedule - need to add to `timetable.test.ts`

tpmcgowan (approved):
LGTM!

PR #505, opened 2023-02-27
VB-1835: Don't allow VisitRestriction to be UNKNOWN on VisitSlot

tpmcgowan:
Minor tweak to improve types. `VisitSlot.visitRestriction` and `visitSessionData.visitRestriction` are used in the context of the booking journey and should only ever be `'OPEN' | 'CLOSED'`. 

A `Visit` can have `visitRestriction` of `UNKNOWN` but this should only be relevant of viewing old migrated visits that can't be updated.


psoleckimoj (approved):
LGTM!

PR #511, opened 2023-03-02
VB-1840 Style location group names as a list

tpmcgowan:
Styling location groups as a single comma separated string looks confusing because the names themselves have commas inside..

This change renders the group names as a list.


AledWEvans (dismissed):
LGTM
Only comment do wonder if class `govuk-!-margin-bottom-1` would be better than `govuk-!-margin-bottom-2`

hutcheonb-moj (approved):
LGTM

tpmcgowan:
> Only comment do wonder if class govuk-!-margin-bottom-1 would be better than govuk-!-margin-bottom-2

Agree, @AledWEvans - I've tweaked the spacing

PR #512, opened 2023-03-02
VB-1881, integration tests for visit by reference

hutcheonb-moj:
## Description
Add integration test for the search for a visit by reference journey.
Add integration test for the search for a visit by prisoner journey.
Added integration tests for the visit details page.


integration_tests/mockApis/visitScheduler.ts
```diff
@@ -203,6 +203,50 @@ export default {
       },
     })
   },
+  stubBookedVisit: (reference: string): SuperAgentRequest => {
+    const result: Visit = {
```
tpmcgowan, line 207: I don't think you actually need this new stub because one already exists for this endpoint (`stubVisit`) - it was added ages ago but isn't used I don't think. So, it could be tidied up and you could just pass in the `Visit` object for it to return rather than just the reference so it's more flexible.

(also, this endpoint on the API can return cancelled visits so `stubVisit` a more accurate name)

integration_tests/integration/searchForABookingReference.cy.ts
```diff
@@ -0,0 +1,72 @@
...
+  beforeEach(() => {
+    cy.task('reset')
+    cy.task('stubSignIn')
+    cy.task('stubAuthUser')
+    cy.task('stubSupportedPrisonIds')
+    cy.task('stubPrisons')
+    cy.signIn()
+  })
+
+  // Check current establishment, change establishment and check again
```
tpmcgowan, line 19: Stray comment and test name need updating 😉

integration_tests/pages/searchForABookingReference.ts
```diff
@@ -0,0 +1,29 @@
+import Page, { PageElement } from './page'
+
+export default class SearchForABookingReferencePage extends Page {
+  constructor() {
+    super('Search for a booking')
+  }
+
+  enterReferenceBlockOne = (number: string): void => {
```
tpmcgowan, line 8: Could perhaps consider something like:
```
enterVisitReference = (reference: string): void => {
  const blocks = reference.split('-')

  cy.get('#searchBlock1').clear().type(blocks[0])
  cy.get('#searchBlock2').clear().type(blocks[1])
  // ...
}
```

then, in your test script, you'd just have a single call: `searchForABookingReferencePage.enterVisitReference('ab-cd-ef-gh')`

tpmcgowan (commented):
Looks good so far!

integration_tests/mockApis/visitScheduler.ts
```diff
@@ -166,20 +166,25 @@ export default {
   stubVisit: (reference: string): SuperAgentRequest => {
     const result: Visit = {
       applicationReference: 'aaa-bbb-ccc',
-      reference: 'ab-cd-ef-gh',
+      reference: 'gh-ef-cd-ab',
```
tpmcgowan, line 169: This works fine, but it would be more flexible if `stubVisit` didn't have hard-coded data. and instead took a visit param (e.g. `stubVisit: (visit: Visit): SuperAgentRequest => ...`) and then returned it (`... jsonBody: visit,`)

In the test script you'd then do something like:
```
const visit = TestData.visit({ // .. any options .. //})
cy.task('stubVisit', visit)
```

More re-usable then; e.g. when moving on to cancellation journey; could stub that with:
```
const cancelledVisit = TestData.visit({ visitStatus: 'CANCELLED' })
cy.task('stubVisit', cancelledVisit)
```

integration_tests/mockApis/prisonerSearch.ts
```diff
@@ -3,14 +3,27 @@ import { stubFor } from './wiremock'
 import { Prisoner } from '../../server/data/prisonerOffenderSearchTypes'
 
 export default {
+  stubPrisoner: (prisoner: Prisoner): SuperAgentRequest => {
+    return stubFor({
+      request: {
+        method: 'GET',
+        url: `/offenderSearch/prison/HEI/prisoners?term=${prisoner.prisonerNumber}`,
```
tpmcgowan, line 10: Could pass `prisonId` as a parameter here (like in `stubPrisoners`) so it's more obvious what's going on when it's called and for more flexibility. Fine for this PR though - just a suggestion for when this next looked at...

integration_tests/pages/searchForBookingByPrisoner.ts
```diff
@@ -0,0 +1,29 @@
+import Page, { PageElement } from './page'
+
+export default class SearchForBookingByPrisonerPage extends Page {
+  constructor() {
+    super('Search for a prisoner')
+  }
+
+  enterVisitReference = (reference: string): void => {
```
tpmcgowan, line 8: Is this mis-named? Perhaps `enterSearchTerm()` or something?

integration_tests/integration/searchForABooking.cy.ts
```diff
@@ -0,0 +1,140 @@
...
+    const searchForBookingByReferencePage = Page.verifyOnPage(SearchForBookingByReferencePage)
+
+    searchForBookingByReferencePage.enterVisitReference('ab-cd-ef-gh')
+
+    cy.task('stubPrisonerById', prisoner)
+    cy.task('stubVisit', visit)
+
+    searchForBookingByReferencePage.continueButton().click()
+
+    searchForBookingByReferencePage.visitReference().contains('ab-cd-ef-gh')
```
tpmcgowan, line 52: I think this is a new `Page` we're on now after the `.click()` - e.g. something like `searchForBookingByReferenceResults`. E.g. the way `pages/searchForAPrisoner.ts` and `pages/searchForAPrisonerResults.ts` are different.

integration_tests/integration/searchForABooking.cy.ts
```diff
@@ -0,0 +1,140 @@
...
+
+    const visitDetailsPage = Page.verifyOnPage(VisitDetailsPage)
+
+    visitDetailsPage.visitReference().contains('ab-cd-ef-gh')
+    visitDetailsPage.prisonerName().contains('Smith, John')
+  })
+
+  it('Should search via prisonerId, than navigate to the summary page', () => {
+    const homePage = Page.verifyOnPage(HomePage)
+    const today = new Date()
```
tpmcgowan, line 70: A lot of these variables and bits of test data are the same for both tests so could be moved up to line 10/11 to avoid duplication

integration_tests/integration/searchForABooking.cy.ts
```diff
@@ -0,0 +1,140 @@
...
+      results: {
+        totalElements: 1,
+        totalPages: 1,
+        content: [prisoner],
+      },
+    })
+
+    searchForBookingByPrisonerPage.continueButton().click()
+
+    searchForBookingByPrisonerPage.resultRow().contains(prisonerDisplayName)
```
tpmcgowan, line 114: Again, after the `.click()` we're now on a different page, e.g. perhaps something like `searchForBookingByPrisonerResultsPage`.

integration_tests/integration/searchForABooking.cy.ts
```diff
@@ -0,0 +1,140 @@
...
+    searchForBookingByPrisonerPage.resultRow().contains('2 April 1975')
+
+    cy.task('stubPrisoner', prisoner)
+    cy.task('stubPrisonerById', prisoner)
+
+    cy.task('stubUpcomingVisits', { offenderNo: prisoner.prisonerNumber, upcomingVisits: [upcomingVisit] })
+
+    searchForBookingByPrisonerPage.prisonerLink().click()
+
+    searchForBookingByPrisonerPage.visitReference().contains('bc-de-fg-hi')
```
tpmcgowan, line 125: This is a different page; perhaps `upcomingVisitsPage` or `visitResultsPage`?

integration_tests/integration/visitDetails.cy.ts
```diff
@@ -0,0 +1,113 @@
...
+    // Additional Information
+    visitDetailsPage.visitComment().contains('Example of a visit comment')
+    visitDetailsPage.visitorConcern().contains('Example of a visitor concern')
+    visitDetailsPage.additionalSupport().contains('Wheelchair ramp, custom request')
+    visitDetailsPage.visitBooked().contains('Monday 14 February 2022 at 10am')
+  })
+
+  it('Should show update/cancel button for future visit', () => {
+    const today = new Date()
+    const prisoner = TestData.prisoner()
```
tpmcgowan, line 64: Variables and set-up common to both tests could be moved to the top of the file to avoid duplication.

tpmcgowan (commented):
Looking good 👍 ! I've added a few comments - mainly just splitting where we havve new `Page`'s and de-duping variables that are common between tests.

tpmcgowan (approved):
LGTM!

PR #514, opened 2023-03-03
VB-1946 Cache tokens in Redis (by using HmppsAuthClient)

tpmcgowan:
Use the default `HmppsAuthClient` client (from the TypeScript template) for requesting system client tokens to leverage the caching it provides in Redis.

Previously, a new token was being requested prior to each API call, often amounting to several token requests per page load. Now the token is requested once for a user and cached for its validity period (~20 mins currently).

On initial log in, a straightforward booking journey now makes one token request rather than ~13.

(Most changes in this PR are actually to update test files. This issue was spotted whilst trying to incorporate changes from https://github.com/ministryofjustice/hmpps-template-typescript/pull/106, so further refactoring and improvements to the tests will be done as part of that.)


psoleckimoj (approved):
LGTM! So basically I was just using it wrong?

PR #516, opened 2023-03-06
Allow specifying Redis system client token name key

psoleckimoj:


tpmcgowan:
Looks good 👍 . A few thoughts:
* I think it would be good to keep the text `systemToken` as part of the key so it's clear what it is when looking at the data (currently sessions are stored as `sess:<random hash value>` and tokens as `systemToken:USERNAME` which is quite clear when searching in redis. 
* Do we actually need to make the token names environment-specific? e.g. would just having, say, `systemToken-staff-ui:` or `systemToken-public-ui:` as key prefixes be sufficient 🤔 ? The environments each have their own Redis instances. Bit torn as I guess we name DBs as a whole differently (e.g. `hwpv-dev`), but not each data item... 🤷‍♂️ 
* Whatever prefixing convention we have, feels like it would be worth adding to the session data key. This is defaulting to `sess:` from `connect-redis` but could be overridden by changing the line in https://github.com/ministryofjustice/book-a-prison-visit-staff-ui/blob/9a00d41a705ba9c04fab6295ed71b3ada1bf2413/server/middleware/setUpWebSession.ts#L18 to something like:
  ```typescript
  store: new RedisStore({ client: client as unknown as Client, prefix: 'sess-custom-prefix:' }),
  ```

helm_deploy/values-dev.yaml
```diff
@@ -23,6 +23,7 @@ generic-service:
     PRISON_REGISTER_API_URL: "https://prison-register-dev.hmpps.service.justice.gov.uk"
     DPS_URL: "https://digital-dev.prison.service.justice.gov.uk/"
     SMS_NOTIFICATIONS_ENABLED: "true"
+    SYSTEM_CLIENT_REDIS_KEY: "bapv-staff-dev"
```
tpmcgowan, line 26: I think this should just be `REDIS_KEY` here and in the other `values-*.yaml` files to match what's in `config.js`

server/config.ts
```diff
@@ -42,6 +42,8 @@ export default {
     port: parseInt(process.env.REDIS_PORT, 10) || 6379,
     password: process.env.REDIS_AUTH_TOKEN,
     tls_enabled: get('REDIS_TLS_ENABLED', 'false'),
+    systemTokenPrefix: `systemToken-${get('REDIS_KEY', 'clientsecretrediskey', requiredInProduction)}:`,
```
tpmcgowan, line 45: Would it be better to default to `bapv-staff` for these two lines instead? That way you'd get the behaviour locally too. I think that would be handy so if you're switching between our set of apps locally they wouldn't be using the same default and thus reusing each other's tokens.

(I guess alternatively could just add the line `REDIS_KEY="bapv-staff"` to the example `.env` config in the `README` 🤷‍♂️)

tpmcgowan (commented):
Looks good - just issue of Helm keys wrong I think? And a comment about default values.

tpmcgowan (approved):
LGTM!

PR #517, opened 2023-03-09
VB-1430 Refactor data access layer dependencies

tpmcgowan:
First step to incorporating the changes in the template app (https://github.com/ministryofjustice/hmpps-template-typescript/pull/106) that move to using a services container and remove the standard router.

This preliminary change refactors data access layer dependencies so they can be used in a subsequent PR to wire up a services container.

Also includes changes from https://github.com/ministryofjustice/hmpps-template-typescript/pull/120 to move `dotenv` to be a dev dependency and ensures app insights is started correctly.


server/data/whereaboutsApiClient.test.ts
```diff
@@ -0,0 +1,44 @@
+import nock from 'nock'
```
tpmcgowan, line 1: Added missing test!

server/data/prisonApiClient.test.ts
```diff
@@ -1,20 +1,25 @@
...
   const token = 'token-1'
 
   beforeEach(() => {
     fakePrisonApi = nock(config.apis.prison.url)
-    client = prisonApiClientBuilder(token)
+    prisonApiClient = new PrisonApiClient(token)
   })
 
   afterEach(() => {
+    if (!nock.isDone()) {
```
tpmcgowan, line 18: Saw this `afterEach()` check on some other projects; looked useful so added to each API client test.

server/data/prisonApiClient.ts
```diff
@@ -7,20 +7,17 @@ import {
...
-
-class PrisonApiClient {
-  constructor(private readonly restclient: RestClient) {}
+  constructor(token: string) {
+    this.restClient = new RestClient('prisonApiClient', config.apis.prison as ApiConfig, token)
+  }
 
-  getBookings(offenderNo: string, prisonId: string): Promise<PagePrisonerBookingSummary> {
-    return this.restclient.get({
+  async getBookings(offenderNo: string, prisonId: string): Promise<PagePrisonerBookingSummary> {
```
tpmcgowan, line 19: Noticed that methods on all our API clients were quite inconsistent as to whether they were marked `async` or not. Added to them all as I assumed they should be 🤔 - does that seem correct?

psoleckimoj: Seems fine to me. I think they're almost not explicitly `async` but actually are through what's calling them so looks fine to me.

psoleckimoj (approved):
LGTM!

PR #522, opened 2023-03-13
VB-1430 Use services container

tpmcgowan:
Follow on from #518, and part of implementing changes in https://github.com/ministryofjustice/hmpps-template-typescript/pull/106.

This change rewires the app to use a `Services` container so dependencies can be passed around more easily and without the need for so much duplication and disruption when services are changed. A lot of changes here, but basically all repetition of the same change to main and corresponding test file.

### DIFF
--- a/server/routes/testutils/appSetup.ts
+++ b/server/routes/testutils/appSetup.ts
@@ -14,27 +14,11 @@ import nunjucksSetup from '../../utils/nunjucksSetup'
 import errorHandler from '../../errorHandler'
 import standardRouter from '../standardRouter'
 import UserService from '../../services/userService'
-import PrisonerSearchService from '../../services/prisonerSearchService'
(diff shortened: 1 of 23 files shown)


server/routes/testutils/appSetup.ts
```diff
@@ -136,107 +104,23 @@ function appSetup({
...
-  app.use(
-    '/visits',
-    visitsRoutes(
-      standardRouter(new MockUserService(), new MockSupportedPrisonsService()),
-      prisonerSearchService,
-      visitSessionsService,
-      auditService,
-    ),
-  )
+  if (services.userService === undefined) {
```
tpmcgowan, line 107: Might look at mocking these two services by default in a different way once changes to `standardRouter` incorporated

psoleckimoj (approved):
LGTM! So much tidier 😁

PR #523, opened 2023-03-13
VB-1430 Simplify mocks set up in tests

tpmcgowan:
Add helper functions to create mock API clients to simplify and avoid repetition in test set up.


hutcheonb-moj (approved):
LGTM

PR #525, opened 2023-03-14
VB-1956/VB-1955/VB-1960/VB-1967 prison config information

hutcheonb-moj:
## Description
Add all the UI prison configuration information for the currently onboarding prisons.
Missing the phone number for Morton Hall.


tpmcgowan (commented):
LGTM - but will wait for missing phone number before approving so we don't accidentally let the `'TBC'` through!

tpmcgowan (approved):
LGTM

PR #532, opened 2023-03-20
WIP - VB-2025, update typescript major version

hutcheonb-moj:
## Description
Update typescript major version
Update all other npm recommendations

Added an exported type for FlashData for all relevant tests

Removed `suppressImplicitAnyIndexErrors` from the `tsconfig.ts`, some methods had to be changed slightly since the type's are now more strict

### DIFF
--- a/server/routes/visitJourney/additionalSupport.test.ts
+++ b/server/routes/visitJourney/additionalSupport.test.ts
@@ -2,15 +2,14 @@ import type { Express } from 'express'
 import request from 'supertest'
 import { SessionData } from 'express-session'
 import * as cheerio from 'cheerio'
-import { VisitSessionData } from '../../@types/bapv'
+import { FlashData, VisitSessionData } from '../../@types/bapv'
 import { appWithAllRoutes, flashProvider } from '../testutils/appSetup'
 import { VisitorSupport } from '../../data/visitSchedulerApiTypes'
 import TestData from '../testutils/testData'
 
 let sessionApp: Express
 
-// Requires bringing in line with the other implementation used, caused issues
-let flashData: Record<'errors' | 'formValues', Record<string, string | string[]>[]>
+let flashData: FlashData
 
 let visitSessionData: VisitSessionData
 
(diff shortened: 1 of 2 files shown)


psoleckimoj (dismissed):
LGTM, could just add `"ignoreDeprecations": "5.0"` to the `.tsignore` file for now? Although I'm happy for this as it's easier to spot issues we need to fix 👍

server/routes/changeEstablishment.ts
```diff
@@ -93,6 +93,7 @@ async function getAvailablePrisonsForUser(
   Object.keys(supportedPrisons)
     .filter(prisonId => userCaseLoadsIds.includes(prisonId))
     .forEach(prisonId => {
+      // @ts-ignore
```
tpmcgowan, line 96: You can remove the `@ts-ignore` and fix the type error in this file by changing line 91 to:
```typescript
const availablePrisonsForUser: Record<string, string> = {}
```

server/routes/visitJourney/additionalSupport.test.ts
```diff
@@ -20,6 +20,7 @@ const availableSupportTypes = TestData.supportTypes()
 beforeEach(() => {
   flashData = { errors: [], formValues: [] }
   flashProvider.mockImplementation(key => {
+    // @ts-ignore
```
tpmcgowan, line 23: You can remove this `@ts-ignore` by changing the line above to:
```typescript
flashProvider.mockImplementation((key: 'errors' | 'formValues') => {
```
So that the index into the  `flashData` object is explicitly typed. Similar fix for most of the other files I think.

(As this is repeated across lots of these test files, we might be able to do something cleverer in future... 🧐)

tpmcgowan (commented):
Had a look at this locally and have added a couple of suggestions for how these TypeScript errors could be fixed. One of them should fix a lot of the errors as it's a repeated error.

Agree with Paul that we could add the ignore option, but as they look pretty fixable and there aren't too many probably worth having a go sorting them...

server/routes/prisoner.test.ts
```diff
@@ -22,9 +22,12 @@ const prisonerSearchService = createMockPrisonerSearchService()
 const visitSessionsService = createMockVisitSessionsService()
 
 const prisonId = 'HEI'
-let flashData: Record<string, string[] | Record<string, string>[]>
+
 let visitSessionData: Partial<VisitSessionData>
 
+type FlashData = Record<string, string[] | Record<string, string>[]>
```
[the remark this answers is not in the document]
hutcheonb-moj: just added to `bapv.d.ts`, think that's the only correct place for it to go?

tpmcgowan (approved):
LGTM!

PR #533, opened 2023-03-22
Update to node18/npm9, redis 7, typescript 5 and others

psoleckimoj:


tsconfig.json
```diff
@@ -16,7 +16,8 @@
     "noImplicitAny": true,
     "experimentalDecorators": true,
     "typeRoots": ["./server/@types", "./node_modules/@types"],
-    "lib": ["es5", "es2015.promise"]
+    "lib": ["es5", "es2015.promise"],
+    "ignoreDeprecations": "5.0"
```
tpmcgowan, line 20: I don't think we need to add this `ignoreDeprecations` as things flagged by that were fixed in Ben's last PR

psoleckimoj: Ahh yes, forgot about this!

psoleckimoj: Removed.

package.json
```diff
@@ -10,7 +10,7 @@
     "compile-sass": "sass --quiet-deps --no-source-map --load-path=node_modules/govuk-frontend --load-path=node_modules/@ministryofjustice/frontend --load-path=. ./assets/sass/application.sass:./assets/stylesheets/application.css --style compressed",
     "watch-ts": "tsc -w",
     "watch-views": "nodemon --watch server/views -e html,njk -x npm run copy-views",
-    "watch-node": "DEBUG=gov-starter-server* nodemon -r dotenv/config --watch dist/ dist/server.js | bunyan -o short",
+    "watch-node": "DEBUG=gov-starter-server* nodemon --watch dist/ dist/server.js | bunyan -o short",
```
psoleckimoj, line 13: @tpmcgowan not sure on this one, I did a pull before starting but didn't change this file 🤔

tpmcgowan: We need the `-r dotenv/config` otherwise it won't load from `.env`. I added this recently when I incorporated changes from https://github.com/ministryofjustice/hmpps-template-typescript/pull/120. Just tried locally and I think this is the case (on a train with patchy signal so hard to be 100% sure 🚂!)

psoleckimoj: Fixed!

tsconfig.json
tpmcgowan: 👍 tbh I don't think it's needed in the template app either - but haven't had time to check yet....

tpmcgowan (approved):
LGTM!

PR #535, opened 2023-03-22
VB-2004 Add session check for additional support

tpmcgowan:
Add check for additional support in the session check middleware, to ensure it is set at the appropriate point in the journey.

(As part of efforts to debug VB-2004, whereby users are occasionally arriving at the 'Check your booking' page with this data not in the session.)


psoleckimoj (approved):
LGTM!

PR #536, opened 2023-03-29
VB-2073 Update dependencies & fix issues identified by CodeQL scan

tpmcgowan:
* Update dependencies
* put `dotenv` back into `devDependencies`
* fix linting errors relating to Cypress
* fix some issues identified by CodeQL scan


integration_tests/pages/additionalSupport.ts
```diff
@@ -8,7 +8,7 @@ export default class AdditionalSupportPage extends Page {
   additionalSupportRequired = (): PageElement => cy.get('[data-test=support-required-yes]')
 
   enterOtherSupportDetails = (details: string): void => {
-    cy.get('#otherSupportDetails').clear().type(details)
```
tpmcgowan, line 11: Updated lint rules for Cypress complained about chaining after `.clear()`. I just removed it because it's currently unnecessary: the inputs _should_ generally be empty - something's gone wrong if they're not. (May need to revisit this in some specific cases - e.g. where things may be pre-populated when testing update journey...)

server/routes/search.ts
```diff
@@ -52,8 +52,8 @@ export default function routes({ auditService, prisonerSearchService, visitSessi
   get(['/prisoner/results', '/prisoner-visit/results'], async (req, res) => {
     const { prisonId } = req.session.selectedEstablishment
     const isVisit = req.originalUrl.includes('-visit')
-    const search = (req.query.search || '') as string
-    const currentPage = (req.query.page || '') as string
+    const search = typeof req.query.search === 'string' ? req.query.search : ''
```
tpmcgowan, line 55: CodeQL flagged these as unsafe because query params can be of type `string` or `string[]`. Just casting it with `as string` placates typescript but doesn't address the actual problem of inadvertently letting an array through when you're just expecting a string.

server/routes/visit.test.ts
```diff
@@ -340,7 +340,7 @@ describe('/visit/:reference', () => {
           expect($('.govuk-back-link').length).toBe(0)
           expect($('[data-test="reference"]').text()).toBe('ab-cd-ef-gh')
 
-          expect(res.text).toContain(`This booking is not for ${supportedPrisons.BLI.replace('&', '&amp;')}`)
+          expect(res.text).toContain(`This booking is not for ${supportedPrisons.BLI.replace(/&/g, '&amp;')}`)
```
tpmcgowan, line 343: Another issue flagged by CodeQL. I didn't realise that `replace` only replaces the first instance in a string. Using a regex with `/g` (global) option does them all. Obviously only a test file and not really an issue here but figured I may as well write it 'properly' 🤓!

PR #539, opened 2023-03-30
VB-2004 Add logging to session middleware check

tpmcgowan:
Add logging of key details when a request fails session check middleware. The aim is to better understand context when these requests fail.

Also remove some obsolete logging.


hutcheonb-moj (approved):
LGTM

PR #540, opened 2023-03-31
VB-1927 Add orchestration service config and switch types from visit scheduler

tpmcgowan:
Putting in place foundations for using new endpoints on the orchestration service:
* Add configuration for the orchestration service
* Update types from all APIs;  add orchestration and remove visit scheduler types
* Switch code to user types from orchestration rather than visit scheduler 
* (Includes a workaround in `server/data/orchestrationApiTypes.ts`for orchestration DTOs having some `enum` properties defined as `string` - VB-2081 filed for this)


psoleckimoj (approved):
LGTM!

PR #541, opened 2023-04-03
VB-2057, request method added to session data and journeys

hutcheonb-moj:
## Description
Incorporated new "Request method" page into booking and update journey
Added request method to `visitSessionData` 
Included in 'Book a visit' integration test journey
`requestMethod` constant file being used
Posts selection, doesn't repopulate on update journey (as is correct)


integration_tests/pages/requestMethod.ts
```diff
@@ -0,0 +1,13 @@
+import Page, { PageElement } from './page'
+
+export default class RequestMethodPage extends Page {
+  constructor() {
+    super('What method was used to make this request?')
+  }
+
+  checkbox1 = (): PageElement => cy.get('#method')
+
+  textValue1 = (): PageElement => cy.get(':nth-child(1) > .govuk-label')
```
tpmcgowan, line 10: Suggestion (untested!) - and you'd need to update the test script...

```suggestion
  textValue = (index: number): PageElement => cy.get(':nth-child(index) > .govuk-label')
```

psoleckimoj (dismissed):
LGTM!

tpmcgowan (commented):
Looks great so far - just need the API building for it now! (Couple of very minor comments added)

integration_tests/pages/requestMethod.ts
```diff
@@ -0,0 +1,13 @@
+import Page, { PageElement } from './page'
+
+export default class RequestMethodPage extends Page {
+  constructor() {
+    super('What method was used to make this request?')
+  }
+
+  checkbox1 = (): PageElement => cy.get('#method')
```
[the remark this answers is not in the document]
hutcheonb-moj: Yeah the first is method, any after get numbered, so `#method` `#method-2`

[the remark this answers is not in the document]
hutcheonb-moj: Yeah the first is method, any after get numbered, so `#method` `#method-2`

integration_tests/pages/requestMethod.ts
hutcheonb-moj: See previous comment, would only work for `textValue` not `checkbox`. 
Also currently only testing the page works, so not using anything other than the first textbox / checkbox

tpmcgowan:
Just tried this out locally, @hutcheonb-moj and it worked really well. Only thing I think could be tweaked is that it doesn't pre-populate the request method form if you go back through the journey from 'check answers' page. I know it's specifically not supposed to pre-populate with previous response when starting an update journey (like other forms do). But, _within_ a book/update journey it feels like it should? This is done on other forms with the `formValues[]` stuff

server/constants/requestMethod.ts
```diff
@@ -0,0 +1,8 @@
+const requestMethods: Record<string, string> = {
```
tpmcgowan, line 1: Now we've got the API definitions in, the `Record<string,string>` could be tied down.

server/views/pages/requestMethod.njk
```diff
@@ -0,0 +1,52 @@
...
+            name: "method",
+            fieldset: {
+                legend: {
+                    text: pageHeaderTitle,
+                    isPageHeading: true,
+                    classes: "govuk-fieldset__legend--l"
+                }
+            },
+            items: methods,
+            errorMessage: errors | findError('method')
```
tpmcgowan, line 41: Should we also have the error summary component on this page?

tpmcgowan (approved):
LGTM

PR #542, opened 2023-04-04
VB-2004 Increase web session timeout to 120 mins

tpmcgowan:
Set web session timeout to 120 mins to reflect template default and be more in line with other apps. Hopefully will reduce number of errors relating to users having expired session.


psoleckimoj (approved):
LGTM!

PR #545, opened 2023-04-05
VB-1927 Don't show user name if NOT_KNOWN

tpmcgowan:
If details of user who booked / updated / created is not available (e.g. old data) then don't display anything. The API returns `NOT_KNOWN` in these cases, so this change catches this value.


psoleckimoj (approved):
LGTM!

PR #550, opened 2023-04-12
VB-1866, PrisonerProfile DTO 

hutcheonb-moj:
## Description
New endpoint being called
Forcing `PrisonerProfile` `visits` array to use @tpmcgowan previous temporary fix:  line 11: `server/data/orchestrationApiTypes.ts` 

This includes part of the work for VB-1468 (Single Visits tab)

Notes:
`VisitBalancesDto` doesn't include the next adjust date, for VO or PVO's.
Possibly new `VisitDto` for returning all visitor names
2 tests skipped for VO count being returned on a REMAND prisoner
1 test skipped for 404 being returned on no prisoner returned

### DIFF
--- a/integration_tests/mockApis/orchestration.ts
+++ b/integration_tests/mockApis/orchestration.ts
@@ -1,6 +1,7 @@
 import { SuperAgentRequest } from 'superagent'
 import { stubFor } from './wiremock'
 import { VisitHistoryDetails } from '../../server/data/orchestrationApiTypes'
+import { PrisonerProfilePage } from '../../server/@types/bapv'
 
 export default {
   stubVisitHistory: (visitHistoryDetails: VisitHistoryDetails): SuperAgentRequest => {
@@ -13,6 +14,27 @@ export default {
         status: 200,
         headers: { 'Content-Type': 'application/json;charset=UTF-8' },
         jsonBody: visitHistoryDetails,
+      },
+    })
+  },
+  stubPrisonerProfile: ({
+    prisonId,
+    prisonerId,
+    profile,
+  }: {
+    prisonId: string
+    prisonerId: string
+    profile: PrisonerProfilePage
+  }): SuperAgentRequest => {
+    return stubFor({
+      request: {
+        method: 'GET',
+        url: `/orchestration/prisoner/${prisonId}/${prisonerId}/profile`,
+      },
+      response: {
+        status: 200,
+        headers: { 'Content-Type': 'application/json;charset=UTF-8' },
+        jsonBody: profile,
       },
     })
   },
--- a/integration_tests/mockApis/orchestration.ts
+++ b/integration_tests/mockApis/orchestration.ts
@@ -1,7 +1,6 @@
 import { SuperAgentRequest } from 'superagent'
 import { stubFor } from './wiremock'
-import { VisitHistoryDetails } from '../../server/data/orchestrationApiTypes'
-import { PrisonerProfilePage } from '../../server/@types/bapv'
+import { PrisonerProfile, VisitHistoryDetails } from '../../server/data/orchestrationApiTypes'
 
 export default {
   stubVisitHistory: (visitHistoryDetails: VisitHistoryDetails): SuperAgentRequest => {
@@ -24,7 +23,7 @@ export default {
   }: {
     prisonId: string
     prisonerId: string
-    profile: PrisonerProfilePage
+    profile: PrisonerProfile
   }): SuperAgentRequest => {
     return stubFor({
       request: {
--- a/server/@types/bapv.d.ts
+++ b/server/@types/bapv.d.ts
@@ -1,4 +1,4 @@
-import { InmateDetail, OffenderRestriction, VisitBalances } from '../data/prisonApiTypes'
+import { OffenderRestriction, VisitBalances } from '../data/prisonApiTypes'
 import { Visit, VisitorSupport, VisitSession } from '../data/orchestrationApiTypes'
 
 export type Prison = {
@@ -138,7 +138,7 @@ export type VisitorListItem = {
   banned: boolean
 }
 
-export type PrisonerProfile = {
+export type PrisonerProfilePage = {
   activeAlerts: PrisonerAlertItem[]
   activeAlertCount: number
   flaggedAlerts: Alert[]
--- a/server/@types/bapv.d.ts
+++ b/server/@types/bapv.d.ts
@@ -139,14 +139,10 @@ export type VisitorListItem = {
 }
 
 export type PrisonerProfile = {
-  displayName: string
-  displayDob: string
   activeAlerts: PrisonerAlertItem[]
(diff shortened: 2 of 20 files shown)


server/data/orchestrationApiTypes.ts
```diff
@@ -27,3 +27,9 @@ export type SessionCapacity = components['schemas']['SessionCapacityDto']
 export type SessionSchedule = components['schemas']['SessionScheduleDto']
 
 export type VisitSession = components['schemas']['VisitSessionDto']
+
+export type Alert = components['schemas']['AlertDto']
+
+export type PrisonerProfile = Omit<components['schemas']['PrisonerProfileDto'], 'visits'> & {
```
tpmcgowan, line 33: If we pull in changes from main (I guess via `develop`) then we no longer need this workaround and the ones above will go to.

server/@types/bapv.d.ts
```diff
@@ -247,3 +242,54 @@ export type VisitsPageSlot = {
...
+  dob: string
+  convictedStatus: 'Convicted' | 'Remand'
+  category: string
+  location: string
+  prisonName: string
+  incentiveLevel: string
+  visitBalances: VisitBalances
+}
+
+export type visitItem = [
```
tpmcgowan, line 258: Would be better as `VisitItem` to keep with case convention on other types.

(I'm assuming this replaces `UpcomingVisitItem` and `PastVisitItem` so those will eventually go? I wonder whether there might be a more compact way to specify this type 🤔 - fine for now though!)

server/data/orchestrationApiClient.ts
```diff
@@ -12,4 +12,8 @@ export default class OrchestrationApiClient {
   async getVisitHistory(reference: string): Promise<VisitHistoryDetails> {
     return this.restClient.get({ path: `/visits/${reference}/history` })
   }
+
+  async getPrisonerProfile(prisonId: string, prisonerId: string): Promise<PrisonerProfile> {
```
tpmcgowan, line 16: Needs a test adding in `orchestrationApiClient.test.ts`

server/services/prisonerProfileService.ts
```diff
@@ -113,17 +93,84 @@ export default class PrisonerProfileService {
...
     })
 
+    const supportedPrisons = await this.supportedPrisonsService.getSupportedPrisons(username)
+
+    const visitsForDisplay: visitItem[] = fullPrisoner.visits.map(visit => {
+      return [
+        {
+          html: `<a href='/visit/${visit.reference}'>${visit.reference}</a>`,
+          attributes: {
+            'data-test': 'tab-upcoming-reference',
```
tpmcgowan, line 103: Is including `tab-upcoming-` in these various attributes here obsolete now that visits are all on one tab? E.g. perhaps prefixing with `tab-visits-` instead?

server/services/prisonerProfileService.test.ts
```diff
@@ -63,26 +67,24 @@ describe('Prisoner profile service', () => {
...
-        latestIepAdjustDate: '2021-04-21',
-        latestPrivIepAdjustDate: '2021-12-01',
-      }
-
-      const pagedVisit: PageVisitDto = {
-        totalPages: 1,
-        totalElements: 1,
-        size: 1,
-        content: [
+      const fullPrisoner = <PrisonerProfile>{
```
tpmcgowan, line 70: We should get a `PrisonerProfile` into the `testData` file for re-use and consistency but that can be done as a subsequent change

server/services/prisonerProfileService.test.ts
```diff
@@ -169,144 +150,127 @@ describe('Prisoner profile service', () => {
...
-            lastName: inmateDetail.lastName,
-            dateOfBirth: inmateDetail.dateOfBirth,
-            agencyId: 'HEI',
-            legalStatus: 'REMAND',
-            convictedStatus: 'Remand',
-          },
-        ],
-        numberOfElements: 1,
+    // Skipped - previously used endpoints were skipped if prisoner was on remand, this logic may wish be to included in the new endpoint
+    it.skip('Does not return visit balances for those on REMAND', async () => {
```
tpmcgowan, line 172: 👍 OK, makes sense to `skip` this for now and we can do some testing to ensure this scenario is handled properly by front-end or orchestration service

server/services/prisonerProfileService.test.ts
```diff
@@ -445,60 +409,62 @@ describe('Prisoner profile service', () => {
...
+            remainingPvo: 2,
+            latestIepAdjustDate: '21 April 2021',
+            latestPrivIepAdjustDate: '1 December 2021',
+          },
+        },
       })
     })
 
-    it('Throws 404 if no bookings found for criteria', async () => {
+    it.skip('Throws 404 if no bookings found for criteria', async () => {
```
tpmcgowan, line 467: We need to test how it handles a 404 from the orchestration service (assuming that that **is** what it returns if a prisoner can't be found). Can sort as a separate PR though.

server/utils/utils.ts
```diff
@@ -184,3 +184,26 @@ export const getWeekOfDatesStartingMonday = (
 
   return { weekOfDates, previousWeek, nextWeek }
 }
+
+export const longPrisonerDateTimePretty = (dateToFormat: string): string => {
```
tpmcgowan, line 188: As discussed, fine to leave these for now if they're working OK but we can probably simplify as a future change because we have other functions / Nunjucks filters that do same / similar.

server/routes/prisoner.test.ts
```diff
@@ -127,10 +152,10 @@ describe('GET /prisoner/A1234BC', () => {
 
         expect($('[data-test="tab-vo-remaining"]').text()).toBe('1')
         expect($('[data-test="tab-vo-last-date"]').text()).toBe('21 April 2021')
-        expect($('[data-test="tab-vo-next-date"]').text()).toBe('15 May 2021')
+        // expect($('[data-test="tab-vo-next-date"]').text()).toBe('15 May 2021')
```
tpmcgowan, line 155: Is orchestration service not returning next VO/PVO allocation date? If so, need to raise a ticket?

tpmcgowan (commented):
Looking great so far - well done! Would sugggest focussing on whatever might be stopping it passing tests in CircleCI (i.e. likely integration tests) and then we can get it into `develop` and tackle the smaller issues - and share them out

integration_tests/mockApis/orchestration.ts
```diff
@@ -16,4 +17,25 @@ export default {
...
     })
   },
+  stubPrisonerProfile: ({
+    prisonId,
+    prisonerId,
+    profile,
+  }: {
+    prisonId: string
+    prisonerId: string
+    profile: PrisonerProfilePage
```
tpmcgowan, line 27: I think the type here should be `PrisonerProfile` from `orchestrationApiTypes` because that is what is returned from the API

PR #551, opened 2023-04-18
VB-2081 Update types and remove workaround for enums as strings

tpmcgowan:
Orchestration service now correctly returns several properties as `enum` rather than `string` so types updated to reflect this and workaround removed.


hutcheonb-moj (approved):
LGTM

PR #555, opened 2023-04-24
VB-2169 Update for 'visitRoom' changing 'capacityGroup'

tpmcgowan:
Incorporate API changes relating to supporting housing levels. Changes use of `visitRoom` to `capacityGroup`.

This data can be `string` or `null`. It isn't used in the front-end but is received with list of visit sessions and passed back when reserving a visit slot.


server/data/visitSchedulerApiClient.test.ts
```diff
@@ -63,33 +62,7 @@ describe('visitSchedulerApiClient', () => {
...
-          {
-            type: 'OTHER',
-            text: 'custom support details',
-          },
-        ],
-        createdBy: 'user1',
-        createdTimestamp: '2022-02-14T10:00:00',
-        modifiedTimestamp: '2022-02-14T10:05:00',
-      }
+      const result = TestData.visit()
```
tpmcgowan, line 65: I've been able to remove some excess test data by using `TestData`. In other places I've left it for another time and just tweaked to avoid unpicking too much.

hutcheonb-moj (approved):
LGTM

PR #558, opened 2023-04-24
VB-1999 Add descriptor to visits timetable page for prisoner category

tpmcgowan:
Update the visits timetable page so that the 'Attendees' column includes prisoner category groups as well as enhanced IEP status and location groups.

### DIFF
--- a/server/views/pages/timetable.njk
+++ b/server/views/pages/timetable.njk
@@ -8,18 +8,28 @@
 
 {% set backLinkHref = "/" %}
 
-{% macro attendees(enhanced, prisonerLocationGroupNames) %}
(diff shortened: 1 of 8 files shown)


server/views/pages/timetable.njk
```diff
@@ -8,18 +8,28 @@
 
 {% set backLinkHref = "/" %}
 
-{% macro attendees(enhanced, prisonerLocationGroupNames) %}
+{% macro attendees(enhanced, prisonerCategoryGroupNames, prisonerLocationGroupNames) %}
```
tpmcgowan, line 11: Trying to wedge several different bits of data in to one table cell here so it's not particularly elegant. I think the design is being reconsidered so hopefully can be simplified in future!

psoleckimoj (approved):
lgtm

PR #561, opened 2023-04-26
VB-2172 Fix session capacity bug when capacity is only 1

tpmcgowan:
Fix bug where available visit slots would not be displayed if the session capacity was set to only one (for open or closed visits).

Also update the wording when no visit slots are available to reflect date/time picker page no longer having filters.


server/services/visitSessionsService.ts
```diff
@@ -111,7 +111,7 @@ export default class VisitSessionsService {
 
         // slotHasCapacity == true when the slot has capacity for this selected restriction
         const slotHasCapacity =
-          visitRestriction === 'OPEN' ? visitSession.openVisitCapacity > 1 : visitSession.closedVisitCapacity > 1
+          visitRestriction === 'OPEN' ? visitSession.openVisitCapacity > 0 : visitSession.closedVisitCapacity > 0
         if (slotHasCapacity) {
```
AledWEvans, line 115: Nice

AledWEvans (approved):
LGTM

PR #563, opened 2023-04-27
VB-1768 Use GDS Tag component for restrictions

tpmcgowan:
Replace the MoJ Pattern Library Badge component with GOV.UK Tag when displaying prisoner/visitor restrictions. Test data example below showing new styles (and default for unhandled restriction codes).

Also:
* tidied up naming of some classes and files
* added `restriction` to `TestData` class
* added some coverage of prisoner / visitor restrictions to Cypress tests

Screenshot 2023-04-27 at 12 42 02


assets/sass/components/_restrictions.scss
```diff
@@ -6,43 +6,43 @@
   }
 }
 
-.visitor-restriction-badge {
+.restriction-tag {
```
tpmcgowan, line 9: Renaming this class (and the file) because these restrictions styles are used for visitors **and** prisoners.

psoleckimoj (approved):
LGTM!

PR #571, opened 2023-05-11
VB-2161 Prisoner profile page - visit balance related fixes

tpmcgowan:
* fix rendering 
* correctly handle visit balances for prisoners on Remand
* code tidy and remove some unused functions/types


server/@types/bapv.d.ts
```diff
@@ -64,16 +64,24 @@ export type PrisonerProfilePage = {
   activeAlerts: PrisonerAlertItem[]
   activeAlertCount: number
   flaggedAlerts: Alert[]
+  prisonerDetails: {
```
tpmcgowan, line 67: Incorporating `PrisonerDetails` and `BAPVVisitBalances` types into `PrisonerProfilePage` for simplicity and because they aren't re-used outside the profile page.

psoleckimoj (approved):
LGTM!

PR #577, opened 2023-05-16
VB-1834: Move visit operations to VisitService

tpmcgowan:
Part of refactoring from using visit scheduler to orchestration service.

* Move visit operations (reserve, change, cancel) to `VisitService`


hutcheonb-moj (approved):
LGTM

PR #580, opened 2023-05-16
VB-1834 Move remaining calls to orchestration client and remove scheduler config

tpmcgowan:
Part of refactoring from using visit scheduler to orchestration service.

* Move remaining calls to `orchestrationApiClient`
* Clear out all visit scheduler related config


infojav (dismissed):

hutcheonb-moj (approved):
LGTM

PR #584, opened 2023-05-22
VB 2245 Vo balance override integration test

hutcheonb-moj:
## Description
Add a check that the VO balance override function on the prisoner profile page is working correctly on the booking journey.


tpmcgowan (approved):
LGTM

PR #588, opened 2023-05-30
Vb 2271 update journey session template reference

hutcheonb-moj:
## Description
Add sessionTemplateReference to the changeReservedVisit call, as this was found to not being updated so was matching visits to the incorrect visit slots.


dhirajshettymoj (approved):
LGTM

PR #590, opened 2023-06-09
VB-1974, added red tag to fully booked visit slots

hutcheonb-moj:
## Description
Added red Gov UK Tag component to fully booked visit sessions on the date and time page.

### DIFF
--- a/server/views/pages/bookAVisit/dateAndTime.test.ts
+++ b/server/views/pages/bookAVisit/dateAndTime.test.ts
@@ -179,10 +179,13 @@ describe('Views - Date and time of visit', () => {
     expect($('#slots-month-March2022-content-1 .bapv-morning-slots > h3').text()).toBe('Morning')
     expect($('#slots-month-March2022-content-1 .bapv-afternoon-slots > h3').eq(1).length).toBe(0) // no afternoon slots
     expect($('label[for="5"]').text()).toContain('9:30am to 10:30am')
-    expect($('label[for="5"]').text()).toContain('Fully booked (30 of 30 tables booked)')
+    expect($('label[for="5"]').text()).toContain('Fully booked')
+    expect($('label[for="5"] .govuk-tag--red').text()).toContain('Fully booked')
+    expect($('label[for="5"]').text()).toContain('(30 of 30 tables booked)')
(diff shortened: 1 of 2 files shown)


server/views/pages/bookAVisit/dateAndTime.test.ts
```diff
@@ -179,10 +179,13 @@ describe('Views - Date and time of visit', () => {
     expect($('#slots-month-March2022-content-1 .bapv-morning-slots > h3').text()).toBe('Morning')
     expect($('#slots-month-March2022-content-1 .bapv-afternoon-slots > h3').eq(1).length).toBe(0) // no afternoon slots
     expect($('label[for="5"]').text()).toContain('9:30am to 10:30am')
-    expect($('label[for="5"]').text()).toContain('Fully booked (30 of 30 tables booked)')
+    expect($('label[for="5"]').text()).toContain('Fully booked')
```
psoleckimoj, line 182: Doesn't this effectively do what line 183 is doing, just less specific?

hutcheonb-moj: because of the way the tag render's additional spacing is inserted, so you'd have to split the line into two (just the .govuk part of the test is superflous), just helps clarify which part is shown inside the tag

hutcheonb-moj: The following is the best you can reduce it to if you don't include the check for tag
   ``` expect($('label[for="6"]').text()).toContain('10:30am to 11:30am')
    expect($('label[for="6"]').text()).toContain('Fully booked')
    expect($('label[for="6"]').text()).toContain('(3 of 1 table booked)') ```

psoleckimoj: Sorry I meant because the selector for both lines will return the same and as we're using `toContain` it's essentially the same. 183 is better as it's more specific and makes sure the tag is there too?

hutcheonb-moj: my misunderstanding completely! amended it

psoleckimoj (approved):
LGTM!

PR #591, opened 2023-06-09
VB-2125, added no approved/permitted warning message

hutcheonb-moj:
## Description
Added no permitted/approved visitors warning banner to select visitors page.
Test added to test for both possible messages
Also changed previous implementation on disabling the checkboxes on this page when the journey isn't able to go ahead

G3094UO - Hewell - All banned visitors

G8316GA - Hewell - Only child visitors

G9472UP - Hewell - No visitors

### DIFF
--- a/server/routes/visitJourney/selectVisitors.test.ts
+++ b/server/routes/visitJourney/selectVisitors.test.ts
@@ -384,6 +384,7 @@ testJourneys.forEach(journey => {
           expect($('input[name="visitors"]').length).toBe(1)
           expect($('[data-test="submit"]').length).toBe(0)
           expect($('[data-test="back-to-start"]').length).toBe(1)
+          expect($('#visitor-4324').attr('disabled')).toBe('disabled')
           expect($('.govuk-warning-text__text').text()).toContain('There are no')
           expect($('.govuk-warning-text__text').text()).toContain('approved')
           expect($('.govuk-warning-text__text').text()).toContain(
@@ -395,7 +396,7 @@ testJourneys.forEach(journey => {
     it('should show back to start button and warning message if only banned visitors listed', () => {
       returnData = [
         {
-          personId: 4324,
+          personId: 3984,
           name: 'John Smith',
           dateOfBirth: '2000-03-02',
           adult: true,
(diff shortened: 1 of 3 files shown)


server/routes/visitJourney/selectVisitors.ts
```diff
@@ -25,6 +25,13 @@ export default class SelectVisitors {
     }
     req.session.visitorList.visitors = visitorList
 
+    let adultVisitors = false
+    visitorList.forEach(visitor => {
```
tpmcgowan, line 29: This will work fine but it will process all the visitors each time - even if the first one it finds is an adult. Admittedly I don't think this is going to be a big performance problem(!) but you could use something like `Array.some()` (see [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some)).

So, (untested) perhaps something like:
```typescript
const adultVisitors = visitorList.some(visitor => visitor.adult === true)
```

(Also, minor point: as `adultVisitors` is a boolean, it might be clearer to name it something like `atLeastOneAdult` or `adultVisitorAvailable` - to avoid confusion with `adultVisitors` elsewhere in the `POST` route where it's used to hold an array of `VisitorListItem`

tpmcgowan (commented):
Looks good - just minor comment about implementation

server/routes/visitJourney/selectVisitors.ts
psoleckimoj, line 29: Would tidier to use something like https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some and the return condition would be what's in your `if` statement.

psoleckimoj: What @tpmcgowan said 😂

hutcheonb-moj: Changed the implementation to include both of these suggestions, thanks!

psoleckimoj (approved):
LGTM!

server/routes/visitJourney/selectVisitors.test.ts
```diff
@@ -397,9 +381,49 @@ testJourneys.forEach(journey => {
...
+        .expect('Content-Type', /html/)
+        .expect(res => {
+          const $ = cheerio.load(res.text)
+          expect($('h1').text().trim()).toBe('Select visitors from the prisoner’s approved visitor list')
+          expect($('[data-test="prisoner-name"]').text()).toBe('John Smith')
+          expect($('input[name="visitors"]').length).toBe(1)
           expect($('[data-test="submit"]').length).toBe(0)
           expect($('[data-test="back-to-start"]').length).toBe(1)
+          expect($('#visitor-3984').attr('disabled')).toBe('disabled')
+          expect($('.govuk-warning-text__text').text()).toContain('There are no')
```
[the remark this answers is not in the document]
hutcheonb-moj: Had to implement the regex method as the whitespace was very persistent, hadn't noticed that method being used in this codebase,, really helps with the clarity thanks!

psoleckimoj (approved):
LGTM!

tpmcgowan (approved):

PR #594, opened 2023-06-15
VB-2345 Don't retry POST or PUT requests

tpmcgowan:
Bring in changes from TypeScript template app to no longer default to retrying requests for `POST` or `PUT`.


psoleckimoj (approved):
LGTM!

PR #598, opened 2023-06-19
Express-Validator major version upgrade

hutcheonb-moj:
## Description
Change express-validator major version - it has changed what's returned as part of the flash data for errors.

Current npm audit warning for - `@aws-sdk/client/sqs` doesn't have a fix, only recommends rolling back the version.

Changed the formError type definition and the implementation of findError and errorSummaryList to accomodate the express-validator changes

Improved all GET test coverage of error messages


psoleckimoj:
@hutcheonb-moj was going to suggest leaving the validator upgrade out of a general update as it's a major update. That way you know all changes are related to that update. Plus it'll need manual flow checks to test the validation all still works properly really.

tpmcgowan (commented):
Looks good so far. In testing locally I have spotted that the error message on prisoner search page isn't linking to the error field correctly.

tpmcgowan (approved):
LGTM!

PR #599, opened 2023-06-21
VB-2376 Temporary fix - don't disable visitor checkboxes (unless banned)

tpmcgowan:
Temporary fix to issue whereby child visitor checkboxes are being wrongly disabled. This fix stops checkboxes being disabled unless the visitor is banned.


hutcheonb-moj (approved):
LGTM

PR #600, opened 2023-06-21
VB-2376, improve selectVisitors to remove bug

hutcheonb-moj:
## Description
Changed implementation of how visitor checkboxes are disabled

Adult has been set to only True or False, being true in the case that no DOB is present. This was for this PR but was also some old logic which needed to be updated

See comments

### DIFF
--- a/server/routes/visitJourney/selectVisitors.ts
+++ b/server/routes/visitJourney/selectVisitors.ts
@@ -25,7 +25,7 @@ export default class SelectVisitors {
     }
     req.session.visitorList.visitors = visitorList
 
-    const atLeastOneAdult = visitorList.some(visitor => visitor.adult === true)
+    const atLeastOneAdult = visitorList.some(visitor => visitor.adult === true || visitor.adult === undefined)
 
     const restrictions = await this.prisonerProfileService.getRestrictions(offenderNo, res.locals.user.username)
     visitSessionData.prisoner.restrictions = restrictions
--- a/server/routes/visitJourney/selectVisitors.ts
+++ b/server/routes/visitJourney/selectVisitors.ts
@@ -25,8 +25,8 @@ export default class SelectVisitors {
     }
     req.session.visitorList.visitors = visitorList
 
-    const atLeastOneAdult = visitorList.some(visitor => visitor.adult === false)
-    const eligibleVisitors = visitorList.some(visitor => visitor.banned === false && visitor.adult === false)
+    const atLeastOneAdult = visitorList.some(visitor => visitor.adult === true)
+    const eligibleVisitors = visitorList.some(visitor => visitor.banned === false && visitor.adult === true)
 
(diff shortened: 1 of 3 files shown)


server/routes/visitJourney/selectVisitors.ts
```diff
@@ -25,7 +25,8 @@ export default class SelectVisitors {
     }
     req.session.visitorList.visitors = visitorList
 
-    const atLeastOneAdult = visitorList.some(visitor => visitor.adult === true)
+    const atLeastOneAdult = visitorList.some(visitor => visitor.adult !== false)
+    const eligibleVisitors = visitorList.some(visitor => visitor.banned === false && visitor.adult !== false)
```
hutcheonb-moj, line 29: moved `eligibleVisitors` to the route as buisness logic shouldn't be on template
Also changed from just checking adult is true, to checking that it's not false as we treat undefined DOB as an adult

tpmcgowan (dismissed):
LGTM

server/routes/visitJourney/selectVisitors.ts
psoleckimoj: What Tom said, there's no need for that change because there is no check for the DOB in this line, adult is either true or false here.

hutcheonb-moj: adult would be undefined if dob is undefined, so just testing for this to be true would mean those visitors wouldn't be counted as an adult

psoleckimoj: `!== false` is exactly the same as `=== true`. It's checking type and value. `adult` is a boolean. Logic needs fixing somewhere else if something is not setting `adult` correctly. ALl you should be doing here is deciding if there is at least one adult.

hutcheonb-moj: `visitorUtils` lines 11 and 12 `buildVisitorListItem` function
```
dateOfBirth: visitor.dateOfBirth,
adult: visitor.dateOfBirth ? isAdult(visitor.dateOfBirth) : undefined,
```

psoleckimoj: Then it's incorrect if you now want an undefined DOB to set adult to true, it should never be undefined if that rule has changed? But that's the right place to set it 🙂

tpmcgowan: OK, I think I understand. Basically:
* for a visitor, `adult` can be `undefined | true | false` (because `adult` is an optional property on `VisitorList`)
* a visitor is counted as 'eligible' if their `adult` 'status' is `undefined | true`

This feels error prone (blame me/Paul as this was there before your time 😅). Feels like we should make `adult` a non-optional property in the type and fix the function that sets it

hutcheonb-moj: Should i make that a seperate PR? it effects 10 files (I've just changed adult to required)

tpmcgowan (approved):
LGTM

PR #601, opened 2023-06-21
Make adult required field, updated tests which had undefined

hutcheonb-moj:
## Description
Change `adult` on `visitorListItem` to be required, in the case that no DOB is returned, adult should be set to true (Based on business logic)


tpmcgowan (approved):
LGTM (although this PR is set to go into `main` - is it supposed to be going into your other PR? Wasn't sure from your comments...)

tpmcgowan (approved):

PR #604, opened 2023-06-27
VB-2357: Handle 'one off' sessions on timetable

tpmcgowan:
On the 'View visits timetable' page, show "One off" for sessions if there is an end date set that matches the session template start date.


infojav (dismissed):

hutcheonb-moj (approved):
LGTM

PR #614, opened 2023-07-10
10th July, minor package updates

hutcheonb-moj:
## Description
Minor package upgrades (Includes gov uk frontend update)

Security errors will continue, only current package change available is rolling back cypress from 12.x to 4.2


psoleckimoj (approved):
LGTM!

PR #615, opened 2023-07-20
VB-2540, incorporate api changes into project

hutcheonb-moj:
## Description
Incorporate API changes for Cost of our service into VSiP, this PR looks to make all journeys functional under these new API's, future PR's will look to add the relevant page and display that information.

Minor note: will need a constants file to display the 'human readable' version of the event history type (currently displays 'BOOKED_VISIT / MIGRATED_VISIT,' instead of Visit booked / Visit migrated, on the booking summary page, next to the time, date, and name of when the action was made

### DIFF
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -30,42 +30,12 @@ export default defineConfig({
         ...auth,
         ...tokenVerification,
 
-        // Orchestration service
-        stubBookVisit: orchestrationService.stubBookVisit,
-        stubCancelVisit: orchestrationService.stubCancelVisit,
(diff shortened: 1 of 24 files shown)


tpmcgowan:
Looks good 👍. Tried out locally and only issue that occurred to me is that the tests don't take into account the fact that there is now a request body when booking - when previously there was none.

In checking out whether this was correct, I basically ended up fixing it 😅. So I've put up https://github.com/ministryofjustice/book-a-prison-visit-staff-ui/pull/616 to merge into this PR if you're happy with that...

tpmcgowan (approved):
LGTM

PR #616, opened 2023-07-20
VB-2540 Update unit and integration tests to include request body for PUT /book

tpmcgowan:
Previously calling the `/book` endpoint didn't require a request body so tests weren't accounting for this. This change adds checks for the `aplicationMethod` on the request body.


hutcheonb-moj (approved):
LGTM

PR #619, opened 2023-07-27
VB-2574 Sort establishment list by prison name

tpmcgowan:
Sort establishment list by prison name, instead of prison code.


hutcheonb-moj (approved):
LGTM

PR #620, opened 2023-07-27
VB-2575 Fix profile tab month headers

tpmcgowan:
Prisoner profile page visits tab month headers show past / upcoming totals - e.g. `August 2023 (1 upcoming)`. If a month **only** had cancelled visit, it would incorrectly render `August 2023 ()`

This fixes it to just show `August 2023` in this case.


hutcheonb-moj (approved):
LGTM

PR #622, opened 2023-07-28
VB-2416 Use updated prisoner profile (containing visitor names)

tpmcgowan:
Use updated `PrisonerProfileDto` that contains visits as `VisitSummaryDto`. These include prison name and visitor names. This avoids having to make an additional calls to look these up separately in the contact registry / prison register.


hutcheonb-moj (approved):
LGTM

PR #624, opened 2023-08-02
Changed 'visit migrated' on summary page to show 'visit booked'

hutcheonb-moj:
## Description
For migrated visit event, the information included is actually about who booked the visit and when they booked the visit, so showing `Visit booked` on the visit summary page is more correct


AledWEvans (approved):
LGTM

PR #625, opened 2023-08-03
Have migrated event audit's show as booked/cancelled instead

hutcheonb-moj:
## Description
For migrated visits - event audit returns `Migrated_visit` as the event type, in this case it's better information for the user if we display Visit booked / Visit cancelled, as the information in the event related to who booked it and the time they did it - not who migrated and when it was done


AledWEvans (approved):
LGTM

PR #634, opened 2023-08-21
VB-1848, Add tabs view on visit details page

hutcheonb-moj:
## Description

Added sub navigation and tabs to prisoner profile page

`govuktimeline` macro wouldn't work so manually built using the example classes

### DIFF
--- a/server/routes/visit.ts
+++ b/server/routes/visit.ts
@@ -64,6 +64,7 @@ export default function routes({
     const reference = getVisitReference(req)
     const fromVisitSearch = (req.query?.from as string) === 'visit-search'
     const fromVisitSearchQuery = req.query?.query as string
+    const selectedTab = req.query?.tab as string
 
     const { visitHistoryDetails, visitors, additionalSupport } = await visitService.getFullVisitDetails({
       reference,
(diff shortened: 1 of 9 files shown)


psoleckimoj (approved):
LGTM!

server/routes/visit.ts
```diff
@@ -64,6 +64,7 @@ export default function routes({
     const reference = getVisitReference(req)
     const fromVisitSearch = (req.query?.from as string) === 'visit-search'
     const fromVisitSearchQuery = req.query?.query as string
+    const selectedTab = req.query?.tab as string
```
tpmcgowan, line 67: Minor point, but I think in general it's safer to actually check types rather than just casting with `as string` (which just basically assumes it's a `string` without actually checking). So, something like:

``` typescript
const selectedTab = typeof req.query.selectedTab === 'string' ? req.query.selectedTab : ''
```

This is because req.query should be treated as untrusted user input - e.g. with meddling it could come though as an array or something which `as string` would just let through.

Appreciate that the lines above do the same thing (probably me when still getting to grips with TypeScript 😅!) - and because you're checking carefully what it is when you use it I think it's OK.

tpmcgowan (approved):
LGTM!

PR #637, opened 2023-08-30
Add attributes for testing

tpmcgowan:
Tweak various `data-test` attributes to make automated testing simpler.


server/views/pages/visit/summary.njk
```diff
@@ -243,24 +241,26 @@
 
           {% if visitHistoryDetails.eventsAudit %}
             <div class="moj-timeline">
+              {% set itemCount = 1 %}
```
tpmcgowan, line 244: Keeping a separate count here because `loop.index` unsuitable as we don't show all items in `eventAudit`.

hutcheonb-moj (approved):
LGTM

PR #638, opened 2023-09-05
VB-1838 Update dependencies and pull in template app changes

tpmcgowan:


docker-compose-local.yml
tpmcgowan, line 1: Deleted - experimental config from quite a while ago that never quite worked

psoleckimoj (approved):
LGTM!

PR #643, opened 2023-09-08
VB-2716 tabs on booking page

hutcheonb-moj:
## Description
Convert visit details page to us MOJ Tabs 

Removing timeline bar if only one event
reducing spacing at top of page
reversing the event order so the most recent event is at the top


tpmcgowan (approved):
LGTM!

PR #645, opened 2023-09-14
VB-2694, show cancel button if less than 28 days after visit start time

hutcheonb-moj:
## Description
Separate cancel/update button show logic
Cancel button displays up to 28 days after the visit start time

### DIFF
--- a/server/routes/visit.test.ts
+++ b/server/routes/visit.test.ts
@@ -329,7 +329,7 @@ describe('/visit/:reference', () => {
         })
     })
 
-    it('should not display update and cancel buttons if start date has passed by one year', () => {
+    it('should not display update and cancel buttons if start date has passed by 29 days', () => {
       const visitDate = new Date(visit.startTimestamp)
       const testDate = visitDate.setDate(visitDate.getDate() + 29)
       jest.useFakeTimers({ advanceTimers: true, now: testDate })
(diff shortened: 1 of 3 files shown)


server/routes/visit.ts
```diff
@@ -99,9 +99,13 @@ export default function routes({
       operationId: res.locals.appInsightsOperationId,
     })
 
+    const oneDayInMs = 1000 * 60 * 60 * 24
```
tpmcgowan, line 102: Minor point but as you're using this as a fixed value, you could put it at the top of the file somewhere instead of inside a route so it's set once rather than being re-calculated for each request. (e.g. similar thing done in `SupportedPrisonService.ts`. 

Similarly, could also make it a bit more explicit/configurable where the '28' days is from. e.g. at the top something like `const CANCELLATION_LIMIT_DAYS=28` . 

The `date-fns` library has various conversion helpers and methods for adding / subtracting / comparing difference days. Don't know if you tried any of these; could well be overkill though I guess...

server/routes/visit.test.ts
```diff
@@ -344,6 +344,21 @@ describe('/visit/:reference', () => {
         })
     })
 
+    it('should display cancel and not the update button if start date has passed by 20 days', () => {
+      const visitDate = new Date(visit.startTimestamp)
+      const testDate = visitDate.setDate(visitDate.getDate() + 20)
```
tpmcgowan, line 349: I think this works OK but for here and the test above I wonder if it's better to test closer to the 28 day boundary. E.g. make this test 27 days and the one above 29? (Sort of ignoring 28 itself as when it 'flips over' will depend on the exact visit time...) Could possibly use `date-fns` `addDays() / subDays()` type methods for readability.

hutcheonb-moj: i found the date-fns add days responded weirdly so kept it as this simple implementation
I did change the tests to 27 and 29 though! Good idea

tpmcgowan (approved):
LGTM

PR #649, opened 2023-09-27
27/09/2023 - update types from orchestration and prison api etc

hutcheonb-moj:
## Description
Update types from Orchestration service and prison api etc

One small change from Orchestration service, just required test data to be changed as it no longer matched on type


tpmcgowan (approved):
LGTM

PR #664, opened 2023-11-03
VB-3177, VB-3182 Fix Date Picker month/year navigation and missing dates

tpmcgowan:
Fix two bugs found in the date picker component:

* bug where navigating between months / years would go to the wrong month. E.g. with 31st October selected, pressing the 'next month' button would take you to December instead of November. The cause was not handling JavaScript dates wrapping if trying to set an invalid date (e.g. in the previous example, 31st November would be invalid so it wrapped to December)
* bug where the month would be truncated to not show the '1st' if this day fell on a Sunday (e.g. Sunday 1st October 2023)

Some Cypress test coverage has been added for these scenarios (ideally to be removed if/when this component is moved to a pattern library)

### DIFF
--- a/assets/js/datepicker.js
+++ b/assets/js/datepicker.js
@@ -474,14 +474,14 @@ Datepicker.prototype.focusLastDayOfWeek = function () {
 Datepicker.prototype.focusNextMonth = function (event, focus = true) {
   event.preventDefault()
   const date = new Date(this.currentDate)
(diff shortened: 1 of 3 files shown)


hutcheonb-moj (dismissed):
LGTM

psoleckimoj (approved):
LGTM!

PR #665, opened 2023-11-06
VB-3181 Use new 'prison names' Prison Register endpoint

tpmcgowan:
Use new endpoint that just returns prison code/name (`/prisons/names`) in place of the existing `/prisons` that returns a lot of unnecessary data. The list of names is still cached within the `SupportedPrisonsService`.

Most of the changes are actually just updating variable names to be more appropriate.


hutcheonb-moj (approved):
LGTM

PR #666, opened 2023-11-07
VB-3188 Tighten search regex for prisoner number

tpmcgowan:
Tighten regex for prisoner number text search to ensure there isn't a partial match within a longer string.


hutcheonb-moj (approved):
LGTM

PR #670, opened 2023-11-14
VB-3115 -set productId to DPS035

dhirajshettymoj:
VB-3115 -set productId to DPS035 and also remove contextColour, v1_2_enabled and v0_47_enabled as these are no longer needed.


tpmcgowan (approved):
LGTM

PR #671, opened 2023-11-15
VB-2648 Allow booking after visitor ban

hutcheonb-moj:
## Description

Allow visitors to make bookings after their ban expires


server/routes/visitJourney/dateAndTime.ts
```diff
@@ -15,17 +16,31 @@ export default class DateAndTime {
   ) {}
 
   async get(req: Request, res: Response): Promise<void> {
+    let daysUntilVisitStart = 2
+
     const isUpdate = this.mode === 'update'
     const { prisonId } = req.session.selectedEstablishment
     const { visitSessionData } = req.session
+
+    let warningBannerText = ''
```
tpmcgowan, line 25: This change will mean that we might potentially have more than one warning message within the same warning component. I don't think this scenario is specified on the ticket, so need to check with Ed/Phoebe what to do in this case. Seems like here all messages will be in one paragraph.

Alternatives might be:
* bulleted individual messages within a single warning component
* multiple warning components; each with a single message

To avoid duplication in the template, an array of warning message strings could be passed, so it could be looped to render either of the above suggestions.

server/data/orchestrationApiClient.ts
```diff
@@ -161,13 +161,13 @@ export default class OrchestrationApiClient {
 
   // orchestration-sessions-controller
 
-  async getVisitSessions(offenderNo: string, prisonId: string): Promise<VisitSession[]> {
+  async getVisitSessions(offenderNo: string, prisonId: string, minNumberOfDays: string): Promise<VisitSession[]> {
```
tpmcgowan, line 164: Might be preferable to make this an optional parameter so we only override defaults when we have to (and thinking ahead to VB-1179)

server/routes/visitJourney/selectVisitors.ts
```diff
@@ -95,12 +96,30 @@ export default class SelectVisitors {
     const closedVisitPrisoner = visitSessionData.prisoner.restrictions.some(
       restriction => restriction.restrictionType === 'CLOSED',
     )
+    // req.session.visitSessionData.earliestDate = getFunctionCalled()
```
tpmcgowan, line 99: Could probably simplify this by something like:
* call  `getBanStatus()` for each visitor to get a `BanStatus[]`
* filter this for where `numDays` is set
* sort this to get the highest value - use this or default (currently 2 days), whichever is greater

server/routes/visitJourney/dateAndTime.ts
```diff
@@ -15,19 +15,33 @@ export default class DateAndTime {
   ) {}
 
   async get(req: Request, res: Response): Promise<void> {
+    let daysUntilVisitStart = 2
+
     const isUpdate = this.mode === 'update'
     const { prisonId } = req.session.selectedEstablishment
     const { visitSessionData } = req.session
+
```
tpmcgowan, line 23: If you use something like this:
```
const warningMessages: { id: string; message: string }[] = []
```
then you have a (default empty) array of warning messsages that you can use with a `for` loop in the template and you'll have the message and id / test values that are necessary

server/routes/visitJourney/selectVisitors.ts
```diff
@@ -96,6 +97,17 @@ export default class SelectVisitors {
       restriction => restriction.restrictionType === 'CLOSED',
     )
 
+    selectedVisitors.forEach(visitor => {
```
tpmcgowan, line 100: Could probably simplyfy this by passing the restrictions for **all visitors** in one go to `getBanStatus()`. E.g. (untested!):
```
const allSelectedVisitorBans: Restriction[] = selectedVisitors.reduce((_acc, visitor) => {
  return visitor.restrictions
}, [])

const banStatus = getBanStatus(allSelectedVisitorBans)
visitSessionData.daysUntilBanExpiry = banStatus.isBanned ? banStatus.numDays : undefined
```

server/routes/visitJourney/selectVisitors.ts
```diff
@@ -96,6 +98,16 @@ export default class SelectVisitors {
       restriction => restriction.restrictionType === 'CLOSED',
     )
 
+    const allSelectedVisitorBans: Restriction[] = selectedVisitors.reduce((acc, visitor) => {
+      visitor.restrictions.forEach(restriction => {
+        acc.push(restriction)
+      })
+      return acc
+    }, [])
+
```
tpmcgowan, line 107: Much simpler solution:
```suggestion
const allSelectedVisitorBans = selectedVisitors.flatMap(visitor => visitor.restrictions)
```
apologies - I think I overcomplicated it in a previous comment...

tpmcgowan (approved):
LGTM

tpmcgowan (approved):
LGTM

PR #681, opened 2023-11-29
VB-3283 Cancellation SMS - handle prisons with no phone number

tpmcgowan:
When sending a cancellation SMS, if a prison doesn't have a contact number use a different Notify template that doesn't include this data.


hutcheonb-moj (approved):
LGTM

PR #683, opened 2023-11-29
VB-3195 Update request methods to include applications made by a prisoner

tpmcgowan:
* Support new "By the prisoner" request method option in the booking journey
* This **is not** included in the list of request methods when cancelling a visit


server/constants/requestMethods.ts
```diff
@@ -1,17 +1,23 @@
 import { ApplicationMethodType } from '../data/orchestrationApiTypes'
 
-export const requestMethodOptions: Readonly<Partial<Record<ApplicationMethodType, string>>> = {
+export const requestMethodsCancellation: Readonly<Partial<Record<ApplicationMethodType, string>>> = {
```
tpmcgowan, line 3: I've made two sets of request methods to support different lists for booking / cancellation

PR #684, opened 2023-12-01
VB-3239 Filter component for bookings to be reviewed listing

tpmcgowan:
Add filter component for the bookings to review listing page. Filter by user name, reason for review - or both.

(Bulk of the filter template and JS/SCSS taken from the https://github.com/ministryofjustice/prepare-a-case repo - with minor customisations.)

### Default view

Screenshot 2023-12-04 at 18 13 57

###  Selecting filters

Screenshot 2023-12-04 at 18 11 55

### With filters selected - showing buttons to clear specific filters or all

Screenshot 2023-12-04 at 18 12 19

### DIFF
--- a/assets/js/filter-component.js
+++ b/assets/js/filter-component.js
@@ -1,0 +1,125 @@
+;(function setupFilterComponent() {
+  'use strict'
+
+  function resizeFilterSections() {
+    var filterButtons = document.getElementsByClassName('pac-filter-button')
+    var filterSections = document.getElementsByClassName('pac-filter-selection')
+    Array.prototype.forEach.call(filterSections, function (element, index) {
+      element.classList.remove('moj-js-hidden')
+      filterButtons[index].style.width =
+        element.offsetWidth > 1 ? element.offsetWidth + 'px' : filterButtons[index].style.width
+      element.classList.add('moj-js-hidden')
+    })
+  }
+
+  function configureFilterButtons() {
+    var filterButtons = document.getElementsByClassName('pac-filter-button')
+    var main = document.getElementsByClassName('govuk-main-wrapper')
+
+    function checkFiltersFocus(e) {
+      var target = e.target && e.target.id.substr(e.target.id.indexOf('-') + 1)
+      var relatedTarget = e.relatedTarget && e.relatedTarget.id.substr(e.relatedTarget.id.indexOf('-') + 1)
+      if (!target || !relatedTarget || relatedTarget.indexOf(target) === -1) {
+        var filterSections = document.getElementsByClassName('pac-filter-selection')
+        Array.prototype.forEach.call(filterButtons, function (element, index) {
+          element.setAttribute('aria-expanded', false)
+          element.classList.remove('pac-filter-button--open')
+          filterSections[index].classList.add('moj-js-hidden')
+          element.active = false
+        })
+      }
+    }
+
+    function toggleFilter(e) {
+      e.stopPropagation()
+      var $el = e.target
+      var current = !$el.active && $el.dataset['controls']
+      var filterSections = document.getElementsByClassName('pac-filter-selection')
+      Array.prototype.forEach.call(filterButtons, function (element, index) {
+        element.setAttribute('aria-expanded', current === element.dataset['controls'])
+        current === element.dataset['controls']
+          ? element.classList.add('pac-filter-button--open')
+          : element.classList.remove('pac-filter-button--open')
+        current === filterSections[index].id
+          ? filterSections[index].classList.remove('moj-js-hidden')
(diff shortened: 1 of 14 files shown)


server/routes/review/visitsReviewListingController.ts
```diff
@@ -12,18 +14,49 @@ export default class VisitsReviewListingController {
   public view(): RequestHandler {
     return async (req, res) => {
       const { prisonName } = req.session.selectedEstablishment
-      const visitsReviewList = await this.visitNotificationsService.getVisitsReviewList(
+      // OK to cast these as express-validator sanitises to string[]
+      const bookedBy = Array.isArray(req.query.bookedBy) ? <string[]>req.query.bookedBy : []
```
tpmcgowan, line 18: Bit of a limitation with `express-validator` is that although its `toArray()` sanitisation function at the bottom will have made these params a string array there doesn't seem to be a way to signal this to TypeScript so you still end up having to convince it we have a `string[]` 🤷‍♂️

server/routes/review/visitsReviewListingController.ts
```diff
@@ -12,18 +14,49 @@ export default class VisitsReviewListingController {
...
+            ...(bookedBy && { bookedBy }),
+            ...(type && { type }),
+          },
+        }),
+      )
+    }
+  }
+
+  public validate(): ValidationChain[] {
+    return [check('bookedBy').toArray(), check('type').toArray()]
```
tpmcgowan, line 60: Only sanitising here; haven't bothered with any _validation_ because list of allowed values depends on the data and invalid values are just discarded when data is filtered; or it'll just show data with no filters applied.

server/services/visitNotificationsService.ts
```diff
@@ -16,32 +19,123 @@ export default class VisitNotificationsService {
...
+
+      const isTypeMatch = appliedFilters.type.length
+        ? appliedFilters.type.some(type => type === notificationGroup.type)
+        : true
+
+      return isUsernameMatch && isTypeMatch
+    })
+  }
+
+  async getVisitsReviewList(
```
tpmcgowan, line 117: A fair bit going on here; now split out to different private helper methods. API just gives all data so identifying potential filter values and actually filtering all done here. 

If in practice there ends up being a lot of data we can revisit and do more work (inc pagination) in the API - but this was discounted for initial release.

assets/js/filter-component.js
```diff
@@ -0,0 +1,118 @@
+;(function setupFilterComponent() {
+  'use strict'
+
+  function resizeFilterSections() {
+    var filterButtons = document.getElementsByClassName('bapv-filter-button')
```
psoleckimoj, line 5: I wonder, could these classname be passed in somehow? I realise there's not much going on with it elsewhere but it'd be nice to make it easier to setup, i.e. just include a file and set some variables instead of trying to find the classnames in the code?

assets/js/filter-component.js
```diff
@@ -0,0 +1,118 @@
...
+        element.offsetWidth > 1 ? element.offsetWidth + 'px' : filterButtons[index].style.width
+      element.classList.add('moj-js-hidden')
+    })
+  }
+
+  function configureFilterButtons() {
+    var filterButtons = document.getElementsByClassName('bapv-filter-button')
+
+    function checkFiltersFocus(e) {
+      var target = e.target && e.target.id.substr(e.target.id.indexOf('-') + 1)
```
psoleckimoj, line 19: `e.target?.id.....` Maybe to tidy up?

assets/sass/components/_filter.scss
```diff
@@ -0,0 +1,144 @@
+.bapv-filter-card {
+  position: relative;
+  background-color: govuk-colour("light-grey");
+  border: 1px solid govuk-colour("mid-grey");
+  padding: 15px 15px 30px 15px;
```
psoleckimoj, line 5: govuk sizing functions? I believe they're in blocks of `5px`

assets/js/filter-component.js
tpmcgowan: Had a look at class names suggestion. Not sure it would ultimately be simpler though. I'm seeing the JS (at assets/js/filter-component.js) and associated Nunjucks template macro (at `server/views/components/filter/template.njk`) as collectively the 'BaPV filter component', so from the main application code you only have to do:
```
{{ bapvFilter({...}) }}
``` 
to use it. I guess it's just the same as the coupling between CSS/JS/HTML in the MoJ / GDS components - but you don't normally see the details as the files are all in `node_modules` 🤷‍♂️!

Have taken on board the point about use of `var` and similar so have freshened up the JS/CSS 🙂.

psoleckimoj (approved):
LGTM!

PR #685, opened 2023-12-04
Add 'data-test' attribute on time slot radios

tpmcgowan:
Add a `data-test` attribute to the radio selections on the data time picker page. This is to make it easier for automated testing. The new attribute contains the date and the start/end times of the slot. For example:

```
<input class="govuk-radios__input" id="7" name="visit-date-and-time" type="radio" value="7" data-test="2023-12-15 - 9am to 10am">
```


AledWEvans (approved):
LGTM

PR #686, opened 2023-12-12
VB-3200-update-type-definition

hutcheonb-moj:
## Description

Import new types (required for knowing the PrisonDTO definition which is used in VB-3200)


tpmcgowan (approved):
LGTM

PR #688, opened 2023-12-12
Acceptance tests - use Docker image with Git pre-installed

tpmcgowan:
Reduce complexity in the CircleCI build by using a `maven` image that already has Git installed so we don't need to install it as an extra job.

(See 'Image variants' on https://hub.docker.com/_/maven for details.)


anunagula3 (approved):
LGTM

PR #689, opened 2023-12-12
VB-3306 Move to hmpps-manage-users-api

tpmcgowan:
Update to use hmpps-manage-users-api instead of hmpps-auth for certain endpoints. Changes copied from typescript template app.

Additionally, because we need the user's `activeCaseLoadId` and this is a deprecated property on the existing endpoint, add support for the Nomis User Roles API to look up this if it is not set.


server/middleware/populateCurrentUser.test.ts
```diff
@@ -0,0 +1,107 @@
...
+import { createMockUserService } from '../services/testutils/mocks'
+import populateCurrentUser from './populateCurrentUser'
+
+const userService = createMockUserService()
+
+let req: Request
+let res: Response
+const next = jest.fn()
+
+describe('populateCurrentUser', () => {
```
tpmcgowan, line 11: Adding test coverage for `populateCurrentUser` middleware because default file from the template changed.

server/middleware/populateCurrentUser.ts
```diff
@@ -8,6 +8,9 @@ export default function populateCurrentUser(userService: UserService): RequestHa
       if (res.locals.user) {
         const user = res.locals.user && (await userService.getUser(res.locals.user.token))
         if (user) {
+          if (!user.activeCaseLoadId) {
```
tpmcgowan, line 11: In practice this shouldn't get called: although undocumented, `getUser` (from manage users API) does in fact return case load. Looking up separately though as advised not to rely on this property being present.

integration_tests/mockApis/manageUsersApi.ts
```diff
@@ -0,0 +1,55 @@
+import { Response } from 'superagent'
+
+import stubUserMe from './nomisUserRoles'
+import { stubFor } from './wiremock'
+
+const stubUser = (name: string) =>
+  stubFor({
+    request: {
+      method: 'GET',
+      urlPattern: '/manage-users-api/users/me',
```
psoleckimoj, line 10: Just need `manage-users-api` changing for consistency with other mockApis

server/middleware/populateCurrentUser.ts
```diff
@@ -8,6 +8,9 @@ export default function populateCurrentUser(userService: UserService): RequestHa
...
+            user.activeCaseLoadId = await userService.getActiveCaseLoadId(res.locals.user.token)
```
psoleckimoj, line 12: `user.token` as you've already got `user` available.

server/services/healthCheck.ts
```diff
@@ -50,6 +50,7 @@ function gatherCheckInfo(aggregateStatus: Record<string, unknown>, currentStatus
 
 const apiChecks = [
   service('hmppsAuth', `${config.apis.hmppsAuth.url}/health/ping`, config.apis.hmppsAuth.agent),
+  service('manageUsersApi', `${config.apis.manageUsersApi.url}/health/ping`, config.apis.manageUsersApi.agent),
   ...(config.apis.tokenVerification.enabled
```
psoleckimoj, line 54: Do you need `nomisUserRolesApi` in here?

tpmcgowan: Yep - good spot! Added all the other APIs we use too as we only had the default ones from the template here.

feature.env
```diff
@@ -3,6 +3,7 @@ HMPPS_AUTH_URL=http://localhost:9091/auth
 HMPPS_AUTH_EXTERNAL_URL=http://localhost:9091/auth
 NOMIS_AUTH_URL=http://localhost:9091/auth
 NOMIS_AUTH_EXTERNAL_URL=http://localhost:9091/auth
+MANAGE_USERS_API_URL=http://localhost:9091/manage-users-api
```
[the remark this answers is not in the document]
tpmcgowan: No reason - just lack of attention on my part. Fixed 😀!

server/middleware/populateCurrentUser.ts
tpmcgowan: Hmm 🤔. Went through it with the debugger to check what's going on here. I think it _does_ have to be `res.locals.user.token` because the `user` here is the result of `userService.getUser()` which is different to the contents of `res.locals.user`. It's only the latter that has the `token`.

psoleckimoj (dismissed):
LGTM!

psoleckimoj (approved):
LGTM!

PR #693, opened 2023-12-14
VB-3295 Visits by date - disable pagination

tpmcgowan:
Temporary fixes to the View visits by date page:
* remove pagination (by hiding prev/next UI and hard-coding max items to 200)
* remove display of adult / child totals 

This quick fix is to enable all visits for a day to be seen on a single page because staff are finding the pagination a hinderance.

There is a new orchestration service endpoint to provide the data for the View visits by date page. The page will be re-written to use this very soon, which will significantly reduce complexity in the front-end - and the pagination code will be 'properly' removed.


psoleckimoj (approved):
LGTM!

PR #695, opened 2023-12-22
VB-3332 Upgrade to GOV.UK Frontend v5 and MoJ Pattern Library v2

tpmcgowan:


assets/sass/components/_prisoner-profile.scss
```diff
@@ -30,6 +30,7 @@
 
   color: white;
   margin-right: govuk-spacing(2);
+  max-width: unset;
```
tpmcgowan, line 33: GOV.UK Tag now has a default `max-width`; removing as it doesn't work well with our use of it.

assets/sass/local.scss
```diff
@@ -1,7 +1,3 @@
-.govuk-main-wrapper a:not(.govuk-button, .govuk-tabs__tab, .bapv-timetable-dates__date--selected>a, .moj-pagination__link, .govuk-error-summary a):visited,
```
tpmcgowan, line 1: Setting class directly on `<a>` tags instead to remove visited state colour; less error-prone.

hutcheonb-moj (approved):
LGTM

hutcheonb-moj (approved):

PR #698, opened 2024-01-05
VB-3200 Fix 'change establishment' error when user has unsupported case load

tpmcgowan:
Fix issue where user arriving in the service (from DPS) with an unsupported caseload would not be able to successfully change to a supported caseload.

### DIFF
--- a/server/routes/changeEstablishment.ts
+++ b/server/routes/changeEstablishment.ts
@@ -59,7 +59,7 @@ export default function routes({ auditService, supportedPrisonsService, userServ
 
     const policyNoticeDaysMin = await supportedPrisonsService.getPolicyNoticeDaysMin(
       res.locals.user.username,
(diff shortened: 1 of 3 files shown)


server/routes/changeEstablishment.ts
```diff
@@ -59,7 +59,7 @@ export default function routes({ auditService, supportedPrisonsService, userServ
 
     const policyNoticeDaysMin = await supportedPrisonsService.getPolicyNoticeDaysMin(
       res.locals.user.username,
-      res.locals.user.activeCaseLoadId,
+      req.body.establishment,
```
tpmcgowan, line 62: This is the bug: on the `POST` to change establishment we should be looking up prison by the newly-selected establishment; **not** the user's current active case load.

hutcheonb-moj (approved):
LGTM

PR #701, opened 2024-01-10
Disable App Insights in staging

tpmcgowan:
With v2.9.2 of the Application Insights client (https://github.com/microsoft/ApplicationInsights-node.js/releases/tag/2.9.2) the app failed in staging because of an invalid key.

App insights isn't set up on staging so this PR unsets the relevant config values to stop it being used.


hutcheonb-moj (approved):
LGTM

PR #707, opened 2024-01-19
Update prisonConfiguration.ts

simonnebMOJ:
Send have requested new rules for the number and age of visitors who can attend their sessions. This needs to be updated in the Manage prison visits UI and go live by 31 January 2024.

The new text should read: 

> You can add up to a total of 4 people, with a maximum of 4 people aged 18 and over. 
> At least one visitor must be 18 or older.


tpmcgowan (approved):
LGTM!

PR #709, opened 2024-01-22
set SMS_NOTIFICATIONS_ENABLED to false on all environments post notify switch to backend.

dhirajshettymoj:
…y switch to backend.


tpmcgowan (approved):
LGTM

PR #711, opened 2024-01-24
VB-3015 Enable pro-active bookings review in pre-prod

tpmcgowan:
* Also, only show details about the two event types being initially released


dhirajshettymoj (approved):
LGTM

PR #713, opened 2024-01-24
VB-2099 Cypress-axe - fail on violations and add ignore list for known accessibility violations

tpmcgowan:
`Cypress-axe` was previously configured to only log accessibility violations. It will now fail a test if there is a violation.

Add a means to define `axeRulesToIgnore` (an array of rule names) in the `Page` test classes so that known accessibility issues (e.g. in GOV.UK components) can be ignored for that particular page.


psoleckimoj (approved):
LGTM!

PR #722, opened 2024-02-14
Update dependencies 2024-02-14 (see notes)

psoleckimoj:
NOTES:

1. `date-fns` major version required an update due to the way a couple of functions work, might be a better way to do it
2. Gov UK kit is now 5.1.0 so has the crown logo or should do.


integration_tests/integration/bookAVisit.cy.ts
```diff
@@ -120,8 +120,8 @@ context('Book a visit', () => {
       cy.contains(contacts[0].restrictions[0].restrictionTypeDescription)
       cy.contains('End date not entered')
     })
-    selectVisitorsPage.getVisitor(contacts[0].personId).check()
-    selectVisitorsPage.getVisitor(contacts[1].personId).check()
+    selectVisitorsPage.getVisitor(contacts[0].personId).check({ force: true })
```
tpmcgowan, line 123: govuk-frontend v5.1.0 has changed something (likely this https://github.com/alphagov/govuk-frontend/pull/4093?) with checkbox input/labels that makes Cypress think a element is hidden when it is not. Fix to force the `.check()`

Not really sure if it's an issue with Cypress or the new markup; but ideally we will be able to remove these overrides at some point....

server/utils/nunjucksSetup.test.ts
```diff
@@ -131,7 +131,7 @@ describe('Nunjucks Filters', () => {
       jest.useRealTimers()
     })
     ;[
-      { input: '2025-11-15', expected: '' }, // erroneous future date of birth
+      { input: '3025-11-15', expected: '' }, // future date of birth
```
tpmcgowan, line 134: Getting in early with a fix to stop this test failing next year 😂!

server/utils/nunjucksSetup.ts
```diff
@@ -94,11 +94,13 @@ export function registerNunjucks(app?: express.Express): Environment {
...
     const duration = intervalToDuration({ start: dob, end: today })
+
     let age = ''
-    if (duration.years < 1) {
-      age = formatDuration(duration, { format: ['months'], zero: true })
-    } else {
+    if (duration.years) {
       age = formatDuration(duration, { format: ['years'] })
+    } else {
+      // workaround below for Duration zero/undefined change (https://github.com/date-fns/date-fns/issues/3658)
```
tpmcgowan, line 102: I've looked into this `date-fns` breaking change a bit more, @psoleckimoj, and re-ordered the code to allow a simpler workaround.

(I've linked to an issue someone else has raised about this. Does seem an odd change, to not distinguish between `undefined` and `0` in the return type 🤷‍♂️)

tpmcgowan (approved):
LGTM

integration_tests/integration/bookAVisit.cy.ts
jsrobertson: Hi @tpmcgowan, just spotted you had to do the same as us [here](https://github.com/ministryofjustice/hmpps-accredited-programmes-ui/pull/504). Just thought I'd mention that I've seen it's been sorted in 5.2.0 now so `force: true` can be removed.

tpmcgowan: Thank-you for letting me know, @jsrobertson  🙂!

PR #725, opened 2024-02-20
VB 3415 - Allow no phone number

hutcheonb-moj:
## Description
Allow no phone number to be an option on the main contact page

Just been added into the journey, no specific integration test for it as we'll soon be re-visiting this code


server/views/pages/bookAVisit/mainContact.njk
```diff
@@ -12,83 +12,118 @@
...
+        }) }}
+        
+        {% set inputPhoneNumberHtml %}
+          {{ govukInput({
+            classes: "govuk-!-width-one-third",
+            label: {
+              text: "Number"
+            },
+            id: "phoneNumberInput",
+            name: "phoneNumberInput",
```
tpmcgowan, line 77: Doesn't really matter, but `id / name` could have remained `phoneNumber` so fewer changes rippling through 🤷‍♂️. I think we've ended up with a field called `phoneNumber` which **isn't** the phone number. 

Perhaps would be clearer if the radios input were named `hasPhoneNumber` with values of `yes | no` and the phone number text input was `phoneNumber`. Lines 19-20 & 53 in `mainContact.ts` might then be easier to understand.

tpmcgowan (commented):
Looks good 👍. Couple of minor suggestions around naming and why integration tests are failing.

Not approving as we need to wait for corresponding back-end changes to be made.

tpmcgowan (approved):
LGTM

PR #726, opened 2024-02-23
VB-3391 Add DPS footer component

tpmcgowan:
Add support for pulling in the DPS frontend components (https://github.com/ministryofjustice/hmpps-micro-frontend-components).

Only the footer component incorporated in this change.

### DIFF
--- a/server/config.ts
+++ b/server/config.ts
@@ -145,6 +145,14 @@ export default {
       },
       agent: new AgentConfig(Number(get('ORCHESTRATION_API_TIMEOUT_RESPONSE', 10000))),
     },
+    frontendComponents: {
+      url: get('COMPONENT_API_URL', 'http://localhost:8082/frontend-components', requiredInProduction),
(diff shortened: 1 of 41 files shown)


server/config.ts
```diff
@@ -145,6 +145,14 @@ export default {
       },
       agent: new AgentConfig(Number(get('ORCHESTRATION_API_TIMEOUT_RESPONSE', 10000))),
     },
+    frontendComponents: {
+      url: get('COMPONENT_API_URL', 'http://localhost:8082/frontend-components', requiredInProduction),
+      timeout: {
+        response: Number(get('FRONTEND_COMPONENTS_TIMEOUT_RESPONSE', 3000)),
```
tpmcgowan, line 151: I've put a fairly low timeout on this so it fails early and renders the fallback.

psoleckimoj (approved):
LGTM!

PR #730, opened 2024-03-05
Change service level implementation when returning additional support

hutcheonb-moj:
## Description

If no additional support set, orchestration doesn't return the `visitorSupport` object, this checks if the object exists before defining in local session data


psoleckimoj (approved):
LGTM!

PR #731, opened 2024-03-06
Fix test attribute

tpmcgowan:
Missing parentheses meant that the `<span>` with the `data-test` attribute wasn't getting generated properly.


hutcheonb-moj (approved):
LGTM

PR #738, opened 2024-03-15
VB-3611 Support 'exclude' location groups on timetable

tpmcgowan:
If location groups on a session template are set to be 'excluded' (rather than the default 'included') then show this on the timetable page.


server/@types/orchestration-api.d.ts
tpmcgowan, line 1: All these files in `@types` are auto generated from Swagger docs so don't particularly need reviewing

dhirajshettymoj (dismissed):
LGTM!

dhirajshettymoj (approved):
LGTM

PR #717, opened 2024-02-05
VB-3296 Rewrite 'Visits by date' page to use new Orchestration service APIs

tpmcgowan:
This is a rewrite of the code behind the 'View visits by date' page. The template/markup of the page is largely unchanged. 

Previously, a lot of work was done in the front-end to get all visits for a day and work back to what the schedules were - as well as looking up prisoner and visitor details. This was complex and slow.

New endpoints on the Orchestration service mean that, for a given day, the scheduled session templates can be found and then visits looked up for these as they are selected. Adult/child counts are no longer needed so no lookups to Contact Registry. Pagination has been removed too, which removes a lot of code.

(As part of this PR there are also some fixes to make the 'Back' link work properly on the booking summary page. It is linked to from several places and was not consistently populating the Back link correctly.)

See also #742 (merged into this) for explanation of handling of 'Unknown' visits.

PR #742, opened 2024-03-25
VB-3296 Visits by date page: handle 'unknown' visits

tpmcgowan:
Building on #717, this adds support for 'unknown' visits to the re-implementation of the View visits by date page. 'unknown' visits means those that have been migrated and therefore don't have a session template reference and may also have a restriction type of `UNKNOWN` instead of `OPEN / CLOSED`.

Because the session schedule and visits-by-template-reference endpoints cannot retrieve these visits, they need to be looked up separately and parsed to construct a navigation based on the time slots that are discovered.

On the page, these are shown under the heading 'All visits', e.g.:

Screenshot 2024-03-25 at 15 43 24

### DIFF
--- a/integration_tests/mockApis/orchestration.ts
+++ b/integration_tests/mockApis/orchestration.ts
@@ -224,7 +224,36 @@ export default {
           prisonCode: { equalTo: prisonId },
           sessionTemplateReference: { equalTo: reference },
           sessionDate: { equalTo: sessionDate },
+          visitStatus: { equalTo: 'BOOKED' },
           visitRestrictions: { equalTo: visitRestrictions },
+        },
+      },
+      response: {
+        status: 200,
+        headers: { 'Content-Type': 'application/json;charset=UTF-8' },
+        jsonBody: visits,
+      },
+    })
+  },
+  stubGetVisitsWithoutSessionTemplate: ({
+    prisonId,
+    sessionDate,
+    visits,
+  }: {
+    prisonId: string
+    sessionDate: string
+    visits: VisitPreview[]
+  }): SuperAgentRequest => {
+    return stubFor({
+      request: {
+        method: 'GET',
+        urlPath: '/orchestration/visits/session-template',
+        queryParameters: {
+          prisonCode: { equalTo: prisonId },
+          sessionTemplateReference: { absent: true },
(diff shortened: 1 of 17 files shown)


integration_tests/mockApis/orchestration.ts
```diff
@@ -234,6 +235,34 @@ export default {
...
+    sessionDate: string
+    visits: VisitPreview[]
+  }): SuperAgentRequest => {
+    return stubFor({
+      request: {
+        method: 'GET',
+        urlPath: '/orchestration/visits/session-template',
+        queryParameters: {
+          prisonCode: { equalTo: prisonId },
+          sessionTemplateReference: { absent: true },
```
tpmcgowan, line 253: Got stuck on this for a while because another mock for the same URL but with a different set of query params was getting matched. Adding the `absent` param means it **won't** match if that param is there - instead of the default of it being optional.

server/routes/visits.ts
```diff
@@ -26,44 +26,71 @@ export default function routes({ auditService, visitService, visitSessionsServic
 
     const { type = '', sessionReference = '', selectedDate = '', firstTabDate = '' } = req.query
 
-    const selectedType: VisitRestriction = type === 'OPEN' || type === 'CLOSED' || type === 'UNKNOWN' ? type : 'OPEN'
+    const selectedSessionReference = sessionReference.toString()
+    const selectedType: VisitRestriction = type === 'OPEN' || type === 'CLOSED' || type === 'UNKNOWN' ? type : undefined
```
psoleckimoj, line 30: Wonder if it might read better as `['OPEN', 'CLOSED', 'UNKNOWN'].includes(type)` Personal preference though.

psoleckimoj (approved):
LGTM!

server/routes/visits.ts
tpmcgowan: Yeah, it's not particularly pretty. I just tried your suggestion which initially looked nicer but it turns out it doesn't actually narrow the type for Typescript 🙁 . It doesn't seem to realise that `type` must now be one of the three values, so complains of type mismatch between `selectedType` and `type`!

psoleckimoj: Go TypeScript 😆

PR #748, opened 2024-04-16
VB-3729 Use the prison's configured maximum total visitors when validating on 'Select visitors' page

tpmcgowan:
Replace the hard-coded maximum of 10 visitors that can be selected for a visit with the configurable value that is now available per prison.

### DIFF
--- a/server/routes/changeEstablishment.ts
+++ b/server/routes/changeEstablishment.ts
@@ -57,7 +57,7 @@ export default function routes({ auditService, supportedPrisonsService, userServ
 
     clearSession(req)
 
-    const policyNoticeDaysMin = await supportedPrisonsService.getPolicyNoticeDaysMin(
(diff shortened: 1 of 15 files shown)


server/routes/changeEstablishment.ts
```diff
@@ -66,6 +66,7 @@ export default function routes({ auditService, supportedPrisonsService, userServ
     const newEstablishment: Prison = {
       prisonId: req.body.establishment,
       prisonName: availablePrisons[req.body.establishment],
+      maxTotalVisitors,
```
tpmcgowan, line 69: Setting this `maxTotalVisitors` in the session like we already do for the booking window (`policyNoticeDaysMin`). A limitation of this is that if the configuration changed a user wouldn't get the new values until the logged in again or changed establishment. Unlikely to be an issue in practice as the configuration will very rarely change.

When we implement the DPS header and take out the current establishment switching code we can address this by getting the prison config at the start of a booking journey.

psoleckimoj (approved):
LGTM!

PR #750, opened 2024-04-22
VB-3760 Hide 'Do not change' button if only notification is a blocked date

tpmcgowan:
If the only visit notification is for a blocked date then do not show the 'Do not change' button on the booking summary page.


hutcheonb-moj (approved):
LGTM

PR #752, opened 2024-04-24
24th April, minor package updates

hutcheonb-moj:
Can't currently fix the reported problem with superagent, as we need supertest to be updated (or it'll continue to import both the new version and old)


tpmcgowan (approved):
LGTM

PR #753, opened 2024-04-24
VB-3762 Update View visits by date page for blocked visit dates

tpmcgowan:
If a date is excluded for visits then on the Visits by date page show a message to this effect (instead of the default "No sessions"). If there are outstanding visits to review then include a link to the Review bookings page.

Also pass correct query parameters to the back link on the Bookings to review page so it can go back as expected in this instance.

### DIFF
--- a/integration_tests/pages/visitsByDate.ts
+++ b/integration_tests/pages/visitsByDate.ts
@@ -29,7 +29,7 @@ export default class VisitsByDatePage extends Page {
 
   prisonerNumber = (index: number): PageElement => cy.get(`:nth-child(${index}) > [data-test="prisoner-number"]`)
 
-  noResultsMessage = (): PageElement => cy.get('#search-results-none')
+  noResultsMessage = (): PageElement => cy.get('[data-test="no-visits-message"]')
 
   // another date form
   toggleChooseAnotherDatePopUp = (): void => {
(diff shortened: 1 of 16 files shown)


integration_tests/pages/visitsByDate.ts
```diff
@@ -29,7 +29,7 @@ export default class VisitsByDatePage extends Page {
 
   prisonerNumber = (index: number): PageElement => cy.get(`:nth-child(${index}) > [data-test="prisoner-number"]`)
 
-  noResultsMessage = (): PageElement => cy.get('#search-results-none')
+  noResultsMessage = (): PageElement => cy.get('[data-test="no-visits-message"]')
```
tpmcgowan, line 32: Tidying up misleading name.

server/data/orchestrationApiClient.ts
```diff
@@ -64,20 +58,6 @@ export default class OrchestrationApiClient {
     return this.restClient.get({ path: `/visits/${reference}/history` })
   }
 
-  async getVisitsByDate(dateString: string, prisonId: string): Promise<PageVisitDto> {
```
tpmcgowan, line 67: Tidying up: we no longer call this endpoint

hutcheonb-moj (approved):
LGTM

PR #755, opened 2024-04-26
VB-3770, VB-3782, VB-3774, VB-3776 Minor API changes and additional support bug fix

tpmcgowan:
Update following API changes:
* Prisons can now be public and/or staff enabled: request `STAFF` prisons
* When creating/updating an application or making this a  booking, pass new flag to `allowOverBooking`

Also, a minor bug fix: when updating a visit that had no additional support added, this choice would not be pre-populated correctly during the update journey.


hutcheonb-moj (approved):
lgtm

PR #761, opened 2024-05-28
VB-3731, update full prisoner visits list url

hutcheonb-moj:
## Description

Change 'View full visit history' button to link to new DPS Prisoner visits page

Added relevant URL's to the config file

Included a test for the URL in integration tests


integration_tests/integration/prisonerProfile.cy.ts
```diff
@@ -106,5 +106,12 @@ context('Prisoner profile page', () => {
       })
     prisonerProfilePage.visitTabVisitStatus().eq(0).contains('Booked')
     prisonerProfilePage.visitTabViewFullHistory().contains('View full visits history')
+    prisonerProfilePage
```
tpmcgowan, line 109: Testing here is OK, but probably worth making sure it's in the unit tests too. E.g. `./server/routes/prisoner.test.ts` has a couple of tests that reference `data-test=view-dps-profile` so could check the URL there.

PR #762, opened 2024-05-28
VB-3731, used wrong url from env files in config

hutcheonb-moj:
## Description

Amend config file to use correct URL


tpmcgowan (approved):
LGTM

PR #764, opened 2024-05-30
Changed SDI visitor guidance

simonnebMOJ:
Now allow max of 4 visitors regardless of age. One must be an adult.


tpmcgowan (approved):
LGTM

PR #765, opened 2024-06-03
Update dependencies 03-06-2024

tpmcgowan:


server/middleware/populateSelectedEstablishment.test.ts
```diff
@@ -90,7 +90,7 @@ describe('populateSelectedEstablishment', () => {
     })
 
     it('should make no changes and not redirect if request path is /change-establishment', async () => {
-      req.path = '/change-establishment'
+      ;(req.path as string) = '/change-establishment'
```
tpmcgowan, line 93: Had to tweak this because eslint seems to have got cleverer and realises `path` in `Express.Request` is `readonly string` so need to override this as just `string`  when mocking it.

psoleckimoj (approved):
LGTM!

PR #766, opened 2024-06-06
VB-3955 Disable submit booking button once clicked

tpmcgowan:
* When the 'Submit booking' button on Check your answers page is clicked, disable it with JavaScript so that it can't be clicked again and potentially cause a duplicate booking
* Add tests for the disabled state to Cypress


integration_tests/pages/checkYourBooking.ts
```diff
@@ -33,5 +33,14 @@ export default class CheckYourBookingPage extends Page {
 
   changeRequestMethod = (): PageElement => cy.get('[data-test="change-request-method"]')
 
-  bookButton = (): PageElement => cy.get('[data-test=submit]')
+  submitBooking = (): void => {
```
tpmcgowan, line 36: Complexity here is because just doing a `.click()` and then checking the button has been disabled won't work because `click()` won't resolve until after the next page has loaded.

psoleckimoj (approved):
LGTM!

PR #767, opened 2024-06-07
VB-3955 Move call to update visit application to earlier in booking journey

tpmcgowan:
Move the call to `/visits/application/{reference}/slot/change` that updates the visit application with accumulated data so that it happens when the 'Main contact' page is submitted instead of the 'Check your answers' page. 

This means that:
* the application is updated as soon as all the required information is available
* the 'Check your details' page now only needs to make a single call (to `/visits/{applicationReference}/book`) so any latency is less likely to result in a user re-trying and ending up making a duplicate request


psoleckimoj (approved):
LGTM!

PR #768, opened 2024-06-12
Move to use CircleCI group in allowlists

psoleckimoj:
Bit of tidying


tpmcgowan (approved):
LGTM - that looks nicer!

PR #775, opened 2024-06-27
VB-4051 / bug fix - actionedBy added to book visit API Call 

hutcheonb-moj:
Add actionedBy to book visit API call

Also uses new eventAudit actionedByFullName and userType


tpmcgowan (approved):
LGTM!

PR #792, opened 2024-08-01
VB-4267, send username when making request to getVisitSesssions

hutcheonb-moj:
Send username when making request to getVisitSessions


tpmcgowan (approved):
LGTM (assuming, @dhirajshettymoj , that back-end is checking username and the prisoner ID for the slot)

PR #793, opened 2024-08-06
VB-3577, prisoner released event

hutcheonb-moj:
Import new eventAuditType and notificationEventType
(2 unused notification types marked as TODO)


integration_tests/integration/reviewAVisit.cy.ts
```diff
@@ -55,6 +30,31 @@ context('Review a visit', () => {
   })
 
   it('should dismiss notifications when a visit is marked as not needing to be updated or cancelled', () => {
+    const eventsAudit: VisitHistoryDetails['eventsAudit'] = [
```
psoleckimoj, line 33: Is there a reason for moving and duplicating these?

server/constants/eventAuditTypes.ts
```diff
@@ -6,6 +6,7 @@ const eventAuditTypes: Partial<Record<EventAuditType, string>> = {
   CANCELLED_VISIT: 'Cancelled',
   MIGRATED_VISIT: 'Migrated',
   NON_ASSOCIATION_EVENT: 'Needs review',
+  PRISONER_RECEIVED_EVENT: 'Neews review',
```
psoleckimoj, line 9: `Neews` ?

server/constants/notificationEvents.ts
```diff
@@ -5,6 +5,9 @@ export const notificationTypes: Readonly<Record<NotificationType, string>> = {
   PRISONER_RELEASED_EVENT: 'Prisoner released',
   PRISONER_RESTRICTION_CHANGE_EVENT: 'Visit type changed',
   PRISON_VISITS_BLOCKED_FOR_DATE: 'Time slot removed',
+  PRISONER_RECEIVED_EVENT: 'Prisoner transfered',
```
psoleckimoj, line 8: Should be `transferred`

integration_tests/integration/reviewAVisit.cy.ts
hutcheonb-moj: was so it's clear at the start of the test which eventsAudit items are appearing in each test, instead of manually changing a particular array numbers entry

psoleckimoj: But everything except the type stays the same? Since it could be set in `beforeEach` you would just have to change `eventsAudit[1].type` on each test which is cleaner.

psoleckimoj (approved):
I'll approve but disagree with moving the `eventsAudit`. @tpmcgowan may want it changing back.

psoleckimoj (approved):
LGTM!

PR #795, opened 2024-08-12
VB-4343, add info endpoint, include active agencies (comes from values.yaml)

hutcheonb-moj:
Add info endpoint (Same implementation as both template and public app)

Include activeAgencies

This comes from values.yaml file, this will need updating when we add prison specific configuration (prisonConfiguration.ts)


psoleckimoj (approved):
LGTM!

PR #796, opened 2024-08-15
VB-4358, replace date picker component

hutcheonb-moj:
Replace manually implemented date picker component with new govuk-frontend component

Quirk found around how it deals with single digit dates now, so changed validation chain to account for new possible outputs from the date picker


integration_tests/integration/datePicker.cy.ts
```diff
@@ -25,6 +25,12 @@ context('Date picker', () => {
   })
 
   it('should navigate through a range of dates and correctly handle month and year boundaries', () => {
+    // Error raised when pressing initial 'toggleChooseAnotherDatePopUp' although it works as intended
```
psoleckimoj, line 28: One worth checking and maybe raising with the MoJ Design Team?

psoleckimoj (approved):
LGTM!

PR #809, opened 2024-09-04
VB-3738, add block dates tile with feature flag

hutcheonb-moj:
Add block dates tile, with feature flag

Feature flag tested in unit tests


server/views/pages/index.njk
```diff
@@ -54,6 +54,17 @@
             "id": "view-timetable"
           }) }}
         </li>
+        {% if showFeatureSessionManagement %}
+          <li class="govuk-grid-column-one-third card-group__item">
+          {{ card({
+            "href": "/date-management",
```
tpmcgowan, line 60:
```suggestion
            "href": "/block-visit-dates",
```
Unless it's specified on the ticket, I'd suggest this alternative URL as GDS guidance is for URLs to align to page titles.

server/routes/index.test.ts
```diff
@@ -46,6 +48,33 @@ describe('GET /', () => {
...
+
+        expect($('[data-test="book-or-change-visit"] .card__link').text()).toBe('Book or change a visit')
+
+        expect($('[data-test="need-review"] .card__link').text()).toContain('Need review')
+
+        expect($('[data-test="view-visits-by-date"] .card__link').text()).toBe('View visits by date')
+
+        expect($('[data-test="view-timetable"] .card__link').text()).toBe('View visits timetable')
+
+        expect($('[data-test="block-dates"] .card__link').text()).toBe('')
```
tpmcgowan, line 76: This is fine but they these are already tested in previous test. So for specifically testing that this card **does not** show then the check of length `4` and an explicit check that the new card is **not** there would be enough: e.g.
```
expect($('[data-test="block-dates"] .card__link').length).toBe(0)
```

tpmcgowan (approved):
LGTM!

PR #811, opened 2024-09-04
VB-3741 Add basic code structure for block visit dates

tpmcgowan:


server/constants/notificationEvents.ts
```diff
@@ -1,13 +1,12 @@
 import { NotificationType } from '../data/orchestrationApiTypes'
 
-export const notificationTypes: Readonly<Record<NotificationType, string>> = {
+// export const notificationTypes: Readonly<Record<NotificationType, string>> = {
```
tpmcgowan, line 3: A few extra types came through here that I don't think we're using. Rather than add more marked `TODO` I've made the type `Partial` so it doesn't complain. Once we support all the available types we can remove this.

server/data/orchestrationApiClient.ts
```diff
@@ -168,6 +168,9 @@ export default class OrchestrationApiClient {
     return this.restClient.get({ path: `/visits/notification/visit/${reference}/types` })
   }
 
+  // orchestration-prisons-exclude-date-controller
+  // TODO put exclude-date calls here
```
tpmcgowan, line 172: Looks like add/remove and get past/future are all in the orchestration service now so we can put them all here 👍

hutcheonb-moj (approved):

server/data/orchestrationApiTypes.ts
```diff
@@ -39,3 +39,5 @@ export type NotificationGroup = components['schemas']['OrchestrationNotification
 export type NotificationVisitInfo = components['schemas']['OrchestrationPrisonerVisitsNotificationDto']
 
 export type PrisonDto = components['schemas']['PrisonDto']
+
+export type PrisonExcludeDateDto = components['schemas']['PrisonExcludeDateDto']
```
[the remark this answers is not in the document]
hutcheonb-moj: while we're working on it - would be easier if it shared the same naming that the backend uses, could re-visit this when we've worked further through this feature

PR #812, opened 2024-09-06
VB-3742 Block a new date for visits page

tpmcgowan:
Render page to block a new visit date (from selected date set in session).

To note:
* actually blocking the date will be in next PR
* date in session will be populated from previous page in journey - being worked on separately

### DIFF
--- a/server/data/orchestrationApiClient.ts
+++ b/server/data/orchestrationApiClient.ts
@@ -11,6 +11,7 @@ import {
   NotificationCount,
   NotificationGroup,
   NotificationType,
+  PageVisitDto,
   PrisonDto,
   PrisonerProfile,
   SessionCapacity,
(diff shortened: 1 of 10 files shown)


server/data/orchestrationApiClient.ts
```diff
@@ -82,6 +83,21 @@ export default class OrchestrationApiClient {
...
+  async getBookedVisitCountByDate(prisonId: string, date: string): Promise<number> {
+    const visits = await this.restClient.get<PageVisitDto>({
+      path: `/visits/search`,
+      query: new URLSearchParams({
+        prisonId,
+        visitStartDate: date,
+        visitEndDate: date,
+        visitStatus: 'BOOKED',
+        page: '0',
+        size: '1',
```
tpmcgowan, line 95: Just using visits `/search` endpoint here to get visit count. Only need to return a single result as that still gives what we need; i.e. `totalElements`.

hutcheonb-moj (approved):

PR #815, opened 2024-09-09
VB-3741 Link block visits list page to add new block page

tpmcgowan:
* Improve listings page (extra options on date picker to hide already-blocked dates; handle no blocked dates)
* 'Wire-up' date picker to POST to the add new block date confirmation page
* Make naming more consistent

(TODO: coming next is the success message when a date is blocked and making the 'unblock' links work)

### DIFF
--- a/integration_tests/mockApis/orchestration.ts
+++ b/integration_tests/mockApis/orchestration.ts
@@ -476,7 +476,7 @@ export default {
       },
     })
   },
-  stubGetFutureExcludeDates: ({
+  stubGetFutureBlockedDates: ({
     prisonId,
     prisonExcludeDates = [TestData.prisonExcludeDateDto()],
   }: {
--- a/server/@types/express/index.d.ts
+++ b/server/@types/express/index.d.ts
@@ -13,7 +13,7 @@ declare module 'express-session' {
     slotsList: VisitSlotList
     visitSessionData: VisitSessionData
     selectedEstablishment: Prison
-    visitBlockDate?: string
+    visitBlockDate?: string // format YYYY-MM-DD
   }
 }
 
@@ -31,6 +31,8 @@ export declare global {
 
       flash(type: 'errors'): ValidationError[]
       flash(type: 'errors', message: ValidationError[]): number
+      flash(type: 'formValues'): Record<string, string | string[] | number[]>[]
+      flash(type: 'formValues', message: Record<string, string | string[] | number[]>): number
     }
 
     interface Locals {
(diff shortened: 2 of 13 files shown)


server/routes/testutils/appSetup.ts
```diff
@@ -122,16 +112,7 @@ function appSetup(
...
-      ...sessionData,
-      regenerate: jest.fn(),
-      destroy: jest.fn(),
-      reload: jest.fn(),
-      id: 'sessionId',
-      resetMaxAge: jest.fn(),
-      save: jest.fn(),
-      touch: jest.fn(),
-    }
+    req.session = sessionData as Session & Partial<SessionData>
```
tpmcgowan, line 115: Improving how `sessionData` is mocked for supertest tests: preserve the object passed in rather than using a new one. Makes it possible to have test assertions for what has changed in the data.

hutcheonb-moj (approved):
lgtm

integration_tests/mockApis/orchestration.ts
```diff
@@ -476,7 +476,7 @@ export default {
       },
     })
   },
-  stubGetFutureExcludeDates: ({
+  stubGetFutureBlockedDates: ({
     prisonId,
     prisonExcludeDates = [TestData.prisonExcludeDateDto()],
```
psoleckimoj, line 481: Do these names match up any more? Future Blocked vs Excluded?

server/@types/express/index.d.ts
```diff
@@ -13,7 +13,7 @@ declare module 'express-session' {
     slotsList: VisitSlotList
     visitSessionData: VisitSessionData
     selectedEstablishment: Prison
-    visitBlockDate?: string
+    visitBlockDate?: string // format YYYY-MM-DD
```
psoleckimoj, line 16: Might be able to be more specific (just found this out though!) 

`type idType = `${number}-${number}-${number}``

psoleckimoj (approved):
Couple of comments but LGTM!

integration_tests/mockApis/orchestration.ts
tpmcgowan: Yeah - all pretty consistently 'block' (matching the actual content/design) in the front-end now rather than 'exclude' . API named differently 🙁

server/@types/express/index.d.ts
tpmcgowan: Hmm... interesting. I didn't know you could do that! Just tried it and it sort of worked - had to replace `string` with `${number}-${number}-${number}` in a few places but it didn't complain. Would probably make sense to define that itself as a type like `SHORT_DATE` I guess. Interesting 🤔!

psoleckimoj: Likewise!

PR #818, opened 2024-09-10
VB-3745 Unblock visit dates

tpmcgowan:
Make 'Unblock' links work to unblock visit dates.

### DIFF
--- a/assets/sass/local.scss
+++ b/assets/sass/local.scss
@@ -291,6 +291,19 @@
   white-space: nowrap;
 }
 
+.bapv-link-button, .bapv-link-button:active, .bapv-link-button:hover {
+  background: none;
+  box-shadow: none;
+  color: $govuk-link-colour;
+  margin: 0;
+  padding: 0;
+  top: unset;
+  text-align: left;
+  @include govuk-link-common;
+  @include govuk-link-style-default;
+}
+
+// TODO fix below - added for Visits by date but affecting other input fields
 .govuk-input__wrapper {
   width: 215px;
   margin-bottom: 24px;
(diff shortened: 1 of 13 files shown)


assets/sass/local.scss
```diff
@@ -291,7 +291,14 @@
   white-space: nowrap;
 }
 
-.govuk-input__wrapper {
-  width: 215px;
-  margin-bottom: 24px;
```
tpmcgowan, line 296: Unrelated - this snuck in to customise layout of date picker on Visits by date page. But, it's too broad and affected other `input` elements. Resolved this with changes above to `bapv-another-date` classes.

assets/sass/local.scss
```diff
@@ -291,7 +291,14 @@
...
+.bapv-link-button, .bapv-link-button:active, .bapv-link-button:hover {
```
tpmcgowan, line 294: Using buttons for the 'Unblock' action so they can submit forms, but styling them to look like links.

psoleckimoj: Hmm, semantically I agree as it's submitting a form but confusing for the user?

psoleckimoj: Yeah, I'd expect a link to go somewhere not submit a form. The other way around, styling a link as a button works fine because it's still an action. Might be worth chatting with design? I have no issue with it looking like a button as that's what it is but it's not my product 😀

tpmcgowan: No worries - I'll raise the issue and see what people say... (tbh I've seen a fair number of DPS things do this so perhaps will be seen as an 'established pattern' for users 🤷‍♂️!)

psoleckimoj: Yep, I seem to remember they had a designer who basically did what they wanted as I've seen it before and was asked if it was OK by another dev.

psoleckimoj (approved):
In which case, LGTM!

PR #822, opened 2024-09-11
VB-3741 Block dates: refactor date validation

tpmcgowan:
* Tweaks to settings for Date Picker component
* Refactor and improve date validation (e.g. reject 31st Feb rather than interpreting it as, e.g. 2nd March)

### DIFF
--- a/server/routes/blockVisitDates/blockVisitDatesController.ts
+++ b/server/routes/blockVisitDates/blockVisitDatesController.ts
@@ -13,7 +13,7 @@ export default class BlockVisitDatesController {
         res.locals.user.username,
       )
 
-      const datePickerMinDate = format(startOfYesterday(), 'dd/MM/yyyy')
+      const datePickerMinDate = format(new Date(), 'dd/MM/yyyy')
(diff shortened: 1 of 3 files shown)


server/routes/blockVisitDates/blockVisitDatesController.ts
```diff
@@ -46,18 +47,18 @@ export default class BlockVisitDatesController {
 
   public validate(): ValidationChain[] {
     return [
+      // store pre-sanitised user-entered value as rawDate to show in field validation error
+      body('rawDate').customSanitizer((_value, { req }) => (typeof req.body?.date === 'string' ? req.body.date : '')),
```
tpmcgowan, line 51: With `express-validator`, once you've run a sanitisation function, you can't access the originally-entered input value. But need to show this back to user with validation messages (GDS error handling guidelines). So, bit of a hack, but setting original value as a property on `req.body` to keep hold of it.

psoleckimoj (approved):
LGTM!

PR #824, opened 2024-09-12
VB-3564 Block visit dates integration tests

tpmcgowan:


integration_tests/components/datePicker.ts
```diff
@@ -0,0 +1,38 @@
+import { PageElement } from '../pages/page'
+
+export default class DatePickerComponent {
```
tpmcgowan, line 3: Moving selector methods for Date Picker to new class so they can be re-used between pages that have the component.

hutcheonb-moj (approved):
LGTM

PR #825, opened 2024-09-12
VB-4397 Use list of prisonIds from API for /info activeAgencies

tpmcgowan:


server/middleware/setUpHealthChecks.ts
```diff
@@ -21,7 +26,15 @@ export default function setUpHealthChecks(applicationInfo: ApplicationInfo): Rou
...
-  router.get('/info', (req, res) => {
+  router.get('/info', async (req, res) => {
+    let activeAgencies: string[]
+    try {
+      activeAgencies = await supportedPrisonsService.getActiveAgencies()
+    } catch (error) {
+      logger.error(error, 'Error retrieving activeAgencies')
+      activeAgencies = []
+    }
+
```
tpmcgowan, line 37: Key part of the change: fetch supported prison IDs from API. Catch errors and handle with empty array. (Did try using `asyncMiddleware` handler but if there's an API error you get HTML error page sent back....

server/services/supportedPrisonsService.ts
```diff
@@ -14,6 +14,10 @@ export default class SupportedPrisonsService {
 
   private lastUpdated = 0
 
+  async getActiveAgencies(): Promise<string[]> {
+    return this.getSupportedPrisonIds(undefined)
```
tpmcgowan, line 18: Calls to `/info` are unauthenticated so there is no `username`. So, call the API without a username which means a system client token gets generated and cached as `%ANONYMOUS%`.

psoleckimoj (approved):
LGTM!

PR #827, opened 2024-09-13
VB-4283, change visits by date page to use exclude date api

hutcheonb-moj:


server/data/orchestrationApiClient.test.ts
```diff
@@ -470,6 +470,22 @@ describe('orchestrationApiClient', () => {
     })
   })
 
+  describe('isBlockedDate', () => {
+    it('should return excludeDateDto if exclude date found for passed date', async () => {
```
tpmcgowan, line 474:
```suggestion
    it('should return boolean indicating whether given date is a blocked date', async () => {
```

server/routes/testutils/testData.ts
```diff
@@ -227,6 +228,14 @@ export default class TestData {
...
   })
 
+  static isExcludeDateDto = ({
+    isExcluded = true,
+    excluded = true,
+  }: Partial<IsExcludeDateDto> = {}): IsExcludeDateDto => ({
+    isExcluded,
+    excluded,
+  })
+
```
tpmcgowan, line 238: I don't think this type is right - I think it's just got one property, `isExcluded`? But, tbh, I don't think it's worth bothering with a `TestData` for it as it is basically just a boolean.

server/data/orchestrationApiClient.test.ts
```diff
@@ -470,6 +470,22 @@ describe('orchestrationApiClient', () => {
...
+      const results = TestData.isExcludeDateDto()
```
tpmcgowan, line 475: Suggest not bothering with this `TestData` method and just putting `true` instead of `results` on line 481

server/services/blockedDatesService.test.ts
```diff
@@ -58,4 +58,28 @@ describe('Blocked dates service', () => {
...
+    it('should return true for given date if exclude date found', async () => {
+      const prisonExcludeDateDto = TestData.prisonExcludeDateDto()
+      const excludedDate = prisonExcludeDateDto.excludeDate
+      orchestrationApiClient.isBlockedDate.mockResolvedValue(true)
+
+      const result = await blockedDatesService.isBlockedDate(prisonId, excludedDate, username)
+
+      expect(orchestrationApiClient.isBlockedDate).toHaveBeenCalledWith(prisonId, excludedDate)
+      expect(result).toStrictEqual(true)
+    })
```
tpmcgowan, line 72: This works but is more complicated than it needs to be; i.e. no need to deal with `prisonExcludeDateDto`s. All that needs testing at this level is that the dependency is called with the correct parameters. So, something like this would do:

```suggestion
    it('should return boolean indicating whether given date is a blocked date', async () => {
      const date = '2000-02-01'
      orchestrationApiClient.isBlockedDate.mockResolvedValue(true)

      const result = await blockedDatesService.isBlockedDate(prisonId, date, username)

      expect(orchestrationApiClient.isBlockedDate).toHaveBeenCalledWith(prisonId, date)
      expect(result).toBe(true)
    })
```
Same for second test (for `false`) - though you could just drop this as one is probably enough because there's no different logic being tested here.

tpmcgowan (commented):
LGTM - haven't been able to approve though as it looks like there's a lint fail and it needs updating with `main`

tpmcgowan (approved):
LGTM

PR #837, opened 2024-09-19
VB-4397 Respond with 503 on /info endpoint if error retrieving data

tpmcgowan:
If there is an error calling the API to get the `activeAgencies` for the `/info` endpoint return an `HTTP 503` response instead of returning normally with `activeAgencies: []`.

Aim is to avoid causing calling services to this endpoint caching incorrect data about the service.


psoleckimoj (approved):
LGTM!

PR #844, opened 2024-09-25
Remove 'uuid' dependency and replace with crypto.randomUUID()

tpmcgowan:
Remove uuid dependency and use Node's crypto.randomUUID() (as advised on https://www.npmjs.com/package/uuid)


psoleckimoj (approved):
LGTM!

PR #848, opened 2024-09-27
VB-4463 allow overbooking end of journey

hutcheonb-moj:


server/views/pages/bookAVisit/overbooking.njk
```diff
@@ -14,36 +15,34 @@
...
+              value: "yes",
+              text: "Yes, book for this timeslot"
+            },
+            {
+              value: "no",
+              text: "No, choose a new date and time"
+            }
+          ],
+          errorMessage: errors | findError('confirmOverbooking'),
+          attributes: { "data-test": "overbooking-confirm" }
```
tpmcgowan, line 37:
```suggestion
          attributes: { "data-test": "confirm-overbooking" }
```
Very minor point - words here the other way round to elsewhere in the code. Doesn't matter which way - just consistency would be better!

server/services/visitSessionsService.ts
```diff
@@ -3,6 +3,7 @@ import { VisitSlot, VisitSlotList, VisitSlotsForDay, VisitSessionData, PrisonerE
 import { VisitSession, SessionCapacity, SessionSchedule } from '../data/orchestrationApiTypes'
 import { ScheduledEvent } from '../data/whereaboutsApiTypes'
 import { HmppsAuthClient, RestClientBuilder, OrchestrationApiClient, WhereaboutsApiClient } from '../data'
+import TestData from '../routes/testutils/testData'
```
tpmcgowan, line 6:
```suggestion
```

server/services/visitSessionsService.test.ts
```diff
@@ -39,6 +39,29 @@ describe('Visit sessions service', () => {
     jest.resetAllMocks()
   })
 
+  describe('getSingleVisitSession', () => {
+    it('should return an array of scheduled sessions for the specified prison and date', async () => {
```
tpmcgowan, line 43:
```suggestion
    it('should return the requested VisitSession', async () => {
```

server/@types/express/index.d.ts
```diff
@@ -13,6 +13,7 @@ declare module 'express-session' {
     slotsList: VisitSlotList
     visitSessionData: VisitSessionData
     selectedEstablishment: Prison
+    allowOverBooking?: boolean
```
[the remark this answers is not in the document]
tpmcgowan: Also, if putting inside `visitSessioData` it could be a **required** rather than optional property so other code can be a bit simpler/tighter.

server/views/pages/bookAVisit/overbooking.njk
```diff
@@ -14,36 +15,34 @@
...
         <span data-test="bookings-count">{{ bookingsCount }}</span> of
         <span data-test="max-capacity">{{ maxCapacity }}</span> {{ "table" | pluralise(maxCapacity) }} are booked for
-        <span data-test="visit-start-time">{{ visitStartTime }}</span> to
-        <span data-test="visit-end-time">{{ visitEndTime }}</span> on 
-        <span data-test="visit-date">{{ visitStartTime | formatDate("EEEE d MMMM") }}</span>.
+        <span data-test="visit-start-time">{{ visitSession.startTimestamp | formatTime }}</span> to
+        <span data-test="visit-end-time">{{ visitSession.endTimestamp | formatTime }}</span> on 
+        <span data-test="visit-date">{{ visitSession.startTimestamp | formatDate("EEEE d MMMM") }}</span>.
       </p>
+      <form action="{{ urlPrefix }}/confirm-overbooking" method="POST" novalidate>
```
tpmcgowan, line 22:
```suggestion
      <form action="{{ urlPrefix }}/check-your-booking" method="POST" novalidate>
```
Could just post directly to where you end up - then don't need the `POST /confirm-overbooking` route defining at all in `./routes/bookAVisit.ts`

server/routes/visitJourney/dateAndTime.ts
```diff
@@ -99,6 +99,9 @@ export default class DateAndTime {
 
     req.session.slotsList = slotsList
 
+    // set allowOverBooking to false on load of select date and time page
+    req.session.allowOverBooking = false
```
tpmcgowan, line 103: if you agree with suggestion to move this inside `visitSessionData` it would instead be set initially in the POST route for prisoner profile (`routes/prisoner.ts`) where `visitSessionData` is defined for a new booking journey.

server/routes/visitJourney/checkYourBooking.ts
```diff
@@ -39,12 +41,23 @@ export default class CheckYourBooking {
...
+    const urlPrefix = getUrlPrefix(isUpdate, visitSessionData.visitReference)
+
+    let { allowOverBooking } = req.session
+
+    if (req.body.confirmOverbooking === 'yes') {
+      allowOverBooking = true
+    } else if (req.body.confirmOverbooking === 'no') {
+      return res.redirect(`${urlPrefix}/select-date-and-time`)
+    }
+
```
tpmcgowan, line 53: I think I understand what's going on here but I think it could be a bit simpler/clearer. E.g., something like

```
const { confirmOverBooking } = req.body  // this will be set if we have come from overbooking confirmation page
if (confirmOverBooking === 'no' ) {
  return res.redirect(`${urlPrefix}/select-date-and-time`)  // i.e. return early if we're going to
}

if (confirmOverBooking === 'yes' ) {
  visitSessionData.allowOverBooking = true
}

// ... and then further down where you call bookVisit(), just pass 'allowOverBooking: visitSessionData.allowOverBooking'
```

server/routes/visitJourney/overbooking.ts
```diff
@@ -0,0 +1,41 @@
...
+import getUrlPrefix from './visitJourneyUtils'
+import { VisitSessionsService } from '../../services'
+
+export default class Overbooking {
+  constructor(
+    private readonly mode: string,
+    private readonly visitSessionsService: VisitSessionsService,
+  ) {}
+
+  async viewFromConfirm(req: Request, res: Response): Promise<void> {
```
tpmcgowan, line 11:
```suggestion
  async get(req: Request, res: Response): Promise<void> {
```
Doesn't really matter, but this matches the naming convention in this part of the app (which I know is different to newer sections / other apps 🙃!)

hutcheonb-moj: I'll leave it as is for now - as I believe my way of implementing the date and time post will have us requiring a different function within this controller - but will wait and see

server/routes/visitJourney/dateAndTime.ts
hutcheonb-moj: having it also reset on load of date time page - as if returning to this page, could be a different session (so should really re-confirm they want to overbook)

server/routes/visitJourney/checkYourBooking.test.ts
```diff
@@ -226,13 +229,39 @@ testJourneys.forEach(journey => {
...
+              allowOverBooking: false,
             })
 
             expect(visitSessionData.visitStatus).not.toBe('BOOKED')
             expect(visitSessionData.visitReference).toBe(journey.isUpdate ? 'ab-cd-ef-gh' : undefined)
             expect(auditService.bookedVisit).not.toHaveBeenCalled()
           })
       })
+
+      describe('Handle API errors', () => {
```
tpmcgowan, line 241: Could include the test above (`'should handle booking failure, display error message and NOT record audit event'`) to this `Handle API errors` block as it's basically just testing for any non-422 API error.

server/routes/visitJourney/dateAndTime.ts
tpmcgowan: 👍 good idea - I hadn't thought of this scenario

PR #850, opened 2024-09-27
VB-4400 Refactor for Prison / PrisonDto

tpmcgowan:
Refactor `getPrisonConfig()` in `SupportedPrisonsService` to be `getPrison()` so all prison config values are available.

Define new `Prison` type to use instead of `PrisonDto` in order to have `prisonId` rather than `code` for consistency.

### DIFF
--- a/server/@types/bapv.d.ts
+++ b/server/@types/bapv.d.ts
@@ -1,8 +1,8 @@
 import { OffenderRestriction } from '../data/prisonApiTypes'
-import { PrisonName } from '../data/prisonRegisterApiTypes'
 import {
   ApplicationMethodType,
   NotificationType,
+  PrisonDto,
   PrisonerProfile,
   Visit,
   VisitorSupport,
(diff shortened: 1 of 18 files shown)


server/@types/bapv.d.ts
```diff
@@ -156,9 +156,8 @@ export type VisitsReviewListItem = {
   visitDates: string[]
 }
 
-export interface Prison extends PrisonName {
-  policyNoticeDaysMin: number
-  maxTotalVisitors: number
+export interface Prison extends Omit<PrisonDto, 'code'> {
```
tpmcgowan, line 159: Change `Prison` to be based on `PrisonDto` (as this has various config values that will be needed), **but** switch property name `code` to `prisonId` for consistency with existing code.

server/services/supportedPrisonsService.ts
```diff
@@ -40,14 +41,10 @@ export default class SupportedPrisonsService {
     return orchestrationApiClient.getSupportedPrisonIds()
   }
 
-  async getPrisonConfig(
-    username: string,
-    prisonCode: string,
-  ): Promise<{ maxTotalVisitors: number; policyNoticeDaysMin: number }> {
+  async getPrison(username: string, prisonId: string): Promise<Prison> {
```
tpmcgowan, line 44: This is the key part of the change: get whole `Prison` rather than just some of the config values.

psoleckimoj (approved):
LGTM!

PR #857, opened 2024-09-30
VB-4400 Use visitor allowance config from API instead of hard-coded values

tpmcgowan:
Remove hard-coded prison configuration data and instead use per-prison data from API.

(Unrelated - fix some integration test errors in logs caused by incorrect Wiremock stubs.)


integration_tests/integration/cancelVisit.cy.ts
```diff
@@ -38,7 +38,7 @@ context('Cancel visit journey', () => {
 
     cy.task('stubPrisonerById', prisoner)
     cy.task('stubVisitHistory', visitHistoryDetails)
-    cy.task('stubPrisonerSocialContacts', { offenderNo, contacts })
+    cy.task('stubPrisonerSocialContacts', { offenderNo, contacts, approvedVisitorsOnly: false })
```
tpmcgowan, line 41: Unrelated to this PR, but here and elsewhere fixed various calls to `stubPrisonerSocialContacts` that have been failing silently (i.e. logging errors but tests still passing) since #799 when a new parameter was introduced.

integration_tests/integration/health.cy.ts
```diff
@@ -51,7 +51,15 @@ context('Healthcheck', () => {
     it('Reports correctly when token verification down', () => {
       cy.task('reset')
       cy.task('stubAuthPing')
+      cy.task('stubManageUsersPing')
```
tpmcgowan, line 54: Fix missing wiremock stub errors caused by health check test.

server/views/pages/bookAVisit/visitors.njk
```diff
@@ -142,9 +142,21 @@
       </p>
       {% endif %}
 
-      {% for textItem in selectVisitorsText %}
-        <p data-test="visitor-information-{{ loop.index }}">{{ textItem }}</p>
-      {% endfor %}
+        <p>
+          Up to <span data-test="visitors-max-total">{{ selectedEstablishment.maxTotalVisitors }} {{ "person" | pluralise(selectedEstablishment.maxTotalVisitors, "people") }}</span>
```
tpmcgowan, line 146: Using same markup has public service Select visitors page.

hutcheonb-moj (approved):

PR #864, opened 2024-10-10
VB-4586 Do not show 'Update visit' button when a visit is flagged because of prisoner transfer/release

tpmcgowan:


hutcheonb-moj (approved):
LGTM

PR #865, opened 2024-10-11
Overbooking page title fix

hutcheonb-moj:


server/views/pages/bookAVisit/overbooking.njk
```diff
@@ -2,7 +2,10 @@
 {% from "govuk/components/radios/macro.njk" import govukRadios %}
 {% from "govuk/components/button/macro.njk" import govukButton %}
 
-{% set pageTitle = "This time slot is fully booked" %}
+{% set pageHeaderTitle = "This time slot is fully booked" %}
+{% set pageTitle %}
+{% if errors | length %}Error: {% endif %}{{ applicationName }} - {{ pageHeaderTitle }}
+{% endset %}
```
tpmcgowan, line 8: At some point we should put these few lines in `layout.njk` like we have in Admin UI - then we don't need to repeat on each page 🤔

tpmcgowan (approved):
LGTM

PR #873, opened 2024-10-23
VB-3751 Use DPS components header

tpmcgowan:
* Replace existing service header with the DPS components header and remove old code
* Remove service phase banner
* Update how populating `selectedEstablishment` is handled
* Minor fixes to tests


integration_tests/e2e/changeEstablishment.cy.ts
```diff
@@ -15,7 +15,7 @@ context('Change establishment', () => {
   })
 
   // Check current establishment, change establishment and check again
-  it('Should change establishment and redirect to home page', () => {
+  it.skip('Should change establishment and redirect to home page', () => {
```
tpmcgowan, line 18: Just skipping this as this whole page will be going in next PR (and it breaks because the links it uses are no longer on the new header)

integration_tests/e2e/signIn.cy.ts
```diff
@@ -37,16 +36,6 @@ context('SignIn', () => {
     Page.verifyOnPage(AuthSignInPage)
   })
 
-  it('User can manage their details', () => {
```
tpmcgowan, line 40: No longer relevant as this service is no longer serving a 'manage details' link.

integration_tests/mockApis/componentApi.ts
```diff
@@ -17,22 +18,30 @@ const meta: HeaderFooterMeta = {
...
+
+const footerHtml = '<footer class="govuk-footer"></footer>'
 
-const stubComponents = () =>
-  stubFor({
+const stubComponents = (userName: string) => {
```
tpmcgowan, line 26: Reworked this to use the actual fallback HTML for the header component. Also allow the username to be set dynamically when setting up the stub so it's closer to what is actually returned from the real component and can be tested.

server/middleware/populateSelectedEstablishment.ts
```diff
@@ -1,15 +1,28 @@
 import type { RequestHandler } from 'express'
 import asyncMiddleware from './asyncMiddleware'
 import { Services } from '../services'
+import logger from '../../logger'
 
 export default function populateSelectedEstablishment({ supportedPrisonsService }: Services): RequestHandler {
   return asyncMiddleware(async (req, res, next) => {
-    if (req.session.selectedEstablishment === undefined && req.path !== '/change-establishment') {
-      const supportedPrisons = await supportedPrisonsService.getSupportedPrisons(res.locals.user.username)
+    if (req.path === '/change-establishment') {
```
tpmcgowan, line 8: Updated this middleware for interim step of using DPS header but still maintaining our own 'Change establishment' page. This is being removed next so this code will get much simpler.

server/views/pages/visit/summary.njk
```diff
@@ -31,7 +31,8 @@
   {% else %}
     {% set backLinkHref = "/prisoner/" + visit.prisonerId %}
   {% endif %}
-
+{% else %}
+  {% set backLinkHref = "/" %}
```
tpmcgowan, line 35: Adding a default back link as otherwise (with new header) there's an edge case where user hits a dead end.

server/utils/nunjucksSetup.ts
```diff
@@ -51,14 +51,7 @@ export function registerNunjucks(app?: express.Express): Environment {
...
 
-  njkEnv.addFilter('initialiseName', (fullName: string) => {
-    // this check is for the authError page
-    if (!fullName) {
-      return null
-    }
-    const array = fullName.split(' ')
-    return `${array[0][0]}. ${array.reverse()[0]}`
-  })
+  njkEnv.addFilter('initialiseName', initialiseName)
```
tpmcgowan, line 54: Moving this filter to be a function in `utils.ts` (in line with how the template app now is) so it can be used in main code as well as Nunjucks files.

hutcheonb-moj (approved):
LGTM

psoleckimoj (approved):
LGTM!

PR #874, opened 2024-10-23
VB-3022 Use DPS header

tpmcgowan:
* Use the standard DPS Header component for the service
* Remove the service's own 'change establishment' feature
* New page to handle case where a user arrives on the service with an unsupported case load
* Refactor to remove several API dependencies that were no longer needed because of the above changes


psoleckimoj (approved):
LGTM!

PR #876, opened 2024-10-24
VB-3754, VB-3755 Remove change establishment page and related code

tpmcgowan:
Remove the 'Change establishment' page and all related code. This will be replaced with a new page to catch cases where a user's case load is not a supported prison.


psoleckimoj (approved):
LGTM!

PR #877, opened 2024-10-25
VB-3753 Handle user having an unsupported active case load

tpmcgowan:
New page to handle user arriving on the service with their case load set to an unsupported establishment. Also triggered if case load changes (e.g. in a different browser window) once the user already has a session.

### DIFF
--- a/integration_tests/mockApis/componentApi.ts
+++ b/integration_tests/mockApis/componentApi.ts
@@ -1,20 +1,14 @@
-import CaseLoad from '@ministryofjustice/hmpps-connect-dps-components/dist/types/CaseLoad'
 import Component from '@ministryofjustice/hmpps-connect-dps-components/dist/types/Component'
 import HeaderFooterMeta from '@ministryofjustice/hmpps-connect-dps-components/dist/types/HeaderFooterMeta'
 import { stubFor } from './wiremock'
 import { convertToTitleCase, initialiseName } from '../../server/utils/utils'
+import TestData from '../../server/routes/testutils/testData'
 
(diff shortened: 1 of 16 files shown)


integration_tests/mockApis/componentApi.ts
```diff
@@ -1,32 +1,24 @@
...
 
 const footerHtml = '<footer class="govuk-footer"></footer>'
 
-const stubComponents = (userName: string) => {
-  const formattedUserName = initialiseName(convertToTitleCase(userName))
+const stubComponents = ({ username, caseLoad }: { username: string; caseLoad: CaseLoad }) => {
```
tpmcgowan, line 12: Allow passing of user case load to components API stub (defaulting in `auth.ts` mock to `HEI` via `TestData.caseLoad()`.

server/views/pages/establishmentNotSupported.njk
```diff
@@ -0,0 +1,19 @@
...
+{% block content %}
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <h1 class="govuk-heading-l">{{ prisonName }} does not use this service</h1>
+
+      <p>Manage prison visits is a new service and only some establishments are using it.</p>
+      
+      <p>
+        If you have access to another establishment that uses this service,
+        click on the link in the DPS header to change your location.
```
psoleckimoj, line 15: Argghh, who decided 'click' was a good word to use 😂 This has been bad accessibility practice since early 2000s.

psoleckimoj (approved):
LGTM!

PR #879, opened 2024-10-25
VB-4789 Remove Prison Register API dependency

tpmcgowan:
* Remove API dependency on Prison Register - this was previously needed to get prison names. No longer necessary as this data now goes from the DPS components meta data and the orchestration service.
* Refactor and simplification of the `populateSelectedEstablishment()` middleware, which essentially checks the user's case load is a supported prison and handles edge cases.


hutcheonb-moj (approved):
LGTM

PR #882, opened 2024-10-31
VB-4787, deal with booker lead visit cancellation

hutcheonb-moj:
Change logic surrounding how the visit history tab gets the 'visit notes' - as booker cancellations don't have notes

Make booker lead cancellations displayed 'cancelled by the visitor' with no reason

### DIFF
--- a/server/views/pages/visit/summary.njk
+++ b/server/views/pages/visit/summary.njk
@@ -47,35 +47,37 @@
     {% set cancelledVisitReason = '' %}
 
       {% if visit.visitStatus === "CANCELLED" %}
-
+      
         {% set status = visit.outcomeStatus.split("_") %}
+
+        {% set cancelledBy %}
+          {% if status[0] === 'ADMINISTRATIVE' %}
+            {{- 'This visit was cancelled due to an administrative error with the booking.' -}}
+          {% elseif status[0] === 'DETAILS' %}
+            {{- 'This visit was cancelled as the details changed after booking.' -}}
+          {% elseif status[0] === 'BOOKER' %}
+            {{- 'This visit was cancelled by the visitor.' -}}
+          {% else %}
+            {{- 'This visit was cancelled by the ' + status[0] | lower + '.' -}}
+          {% endif %}
+        {% endset %}
+
+        {% set cancelledVisitBannerHtml %}
+          {{ govukWarningText({
+            classes: "govuk-!-margin-bottom-0 govuk-!-padding-left-4",
+            html: '<span data-test="visit-cancelled-type">' + cancelledBy + '</span>',
+            iconFallbackText: "Warning"
+          }) }}
+        {% endset %}
+
+        {{ mojBanner({
+          classes: "govuk-!-padding-left-4",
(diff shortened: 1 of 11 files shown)


server/views/pages/visit/summary.njk
```diff
@@ -47,35 +47,37 @@
...
-              classes: "govuk-!-padding-left-4",
-              html: cancelledVisitBannerHtml
-            }) }}
+        {% set cancelledBy %}
+          {% if status[0] === 'ADMINISTRATIVE' %}
+            {{- 'This visit was cancelled due to an administrative error with the booking.' -}}
+          {% elseif status[0] === 'DETAILS' %}
+            {{- 'This visit was cancelled as the details changed after booking.' -}}
+          {% elseif status[0] === 'BOOKER' %}
+            {{- 'This visit was cancelled by the visitor.' -}}
```
psoleckimoj, line 59: Shouldn't this be the booker?

hutcheonb-moj: https://dsdmoj.atlassian.net/browse/VB-4787

No Ed asked specifically that is said visitor

server/views/pages/visit/summary.njk
```diff
@@ -47,35 +47,37 @@
     {% set cancelledVisitReason = '' %}
 
       {% if visit.visitStatus === "CANCELLED" %}
-
+
```
psoleckimoj, line 50: Added white space?

server/views/pages/visit/summary.njk
```diff
@@ -47,35 +47,37 @@
...
+            iconFallbackText: "Warning"
+          }) }}
+        {% endset %}
+
+        {{ mojBanner({
+          classes: "govuk-!-padding-left-4",
+          html: cancelledVisitBannerHtml
+        }) }}
+
+        {%- for visitNote in visit.visitNotes %}
```
psoleckimoj, line 78: Is there a reason this has been moved?

hutcheonb-moj: The page can't work without moving how this section was done

visitNotes don't exist for booker lead visits, so it was blocking the banner from displaying completely

Logically this section shouldn't have been tied together like it had - as the visitNote loop section is only used to set the `cancelledVisitReason` (which is then used in the visit history details tab)

psoleckimoj: It could probably get done in the route rather than in the template maybe as you've got a lot more control with the functions etc available that doing it here?

psoleckimoj: Just pass in as an extra arg.

tpmcgowan: Discussed this with Ben and agreed that too much going on in template here. We've filed VB-4859 to fix this by taking this logic out and putting it in a utility function or similar so it can be more easily tested too.

tpmcgowan (dismissed):
LGTM

tpmcgowan (dismissed):

tpmcgowan (approved):
LGTM

PR #888, opened 2024-11-19
VB-4331 Handle main contact email address

tpmcgowan:
Add support for visits originating in the public service which have a main contact email address.

* Booking details page: show main contact email/phone - or a message if no contact details
* Update journey: although email isn't collected in the staff booking journey, ensure it is retained if a visit originating in the public service is updated


server/routes/visit.test.ts
```diff
@@ -211,41 +212,19 @@ describe('/visit/:reference', () => {
         })
     })
 
-    it('should render full booking summary page with visit information and prisoner tab selected, with default back link, formatting unknown contact telephone correctly', () => {
-      visitHistoryDetails.visit.visitContact.telephone = 'UNKNOWN'
```
tpmcgowan, line 215: Checked with Dhiraj and we don't need to handle `UNKNOWN` as a special value for `telephone` so tidying this up.

hutcheonb-moj (approved):
LGTM

PR #889, opened 2024-11-21
VB-4655 Prisoner alerts: render newlines in alert descriptions

tpmcgowan:
Preserve line spaces when rendering alert descriptions so user formatting is retained.


hutcheonb-moj (approved):
LGTM

PR #900, opened 2024-12-17
VB-3017, add link to all prisoner alerts, with new content

hutcheonb-moj:
Add new content to prisoner profile page, include link to all alerts


server/routes/prisoner.ts
```diff
@@ -28,10 +29,13 @@ export default function routes({ auditService, prisonerProfileService }: Service
       operationId: res.locals.appInsightsOperationId,
     })
 
+    const prisonerDpsAlertsUrl = `${config.dpsPrisoner}prisoner/${prisonerId}/alerts/active`
```
tpmcgowan, line 32: Could just build this URL in the template like we do for 'full visit history' button to avoid creating here and passing through 🤷‍♂️. This is fine though

PR #905, opened 2024-12-24
VB-3644 page titles

hutcheonb-moj:
Fix page title issues

Also address some bad tab spacing on particular pages

Also tackle VB-4728, opted for 'applicationName' as full title

### DIFF
--- a/server/views/pages/visit/summary.njk
+++ b/server/views/pages/visit/summary.njk
@@ -18,8 +18,7 @@
   {% endif %}
 {% endmacro %}
 
-{% set pageHeaderTitle = "Visit booking details" %}
-{% set pageTitle = pageHeaderTitle + " (" + visit.reference + ") - " + applicationName %}
+{% set pageHeaderTitle = "Visit booking details (" + visit.reference + ")" %}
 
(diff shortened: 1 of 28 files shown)


server/views/pages/visit/summary.njk
```diff
@@ -19,7 +19,7 @@
 {% endmacro %}
 
 {% set pageHeaderTitle = "Visit booking details" %}
-{% set pageTitle = applicationName + " - " + pageHeaderTitle + " - " + visit.reference %}
+{% set pageTitle = pageHeaderTitle + " - " + applicationName + " - " + visit.reference %}
```
tpmcgowan, line 22:
```suggestion
{% set pageTitle = pageHeaderTitle + " (" + visit.reference + ") - " + applicationName  %}
```
Not sure if this is right as you get a title like `Visit booking details - Manage prison visits - za-lc-ev-mv - DPS`.

Suggestion would render as: `Visit booking details (za-lc-ev-mv) - Manage prison visits - DPS` 🤷‍♂️

tpmcgowan (approved):
LGTM!

PR #910, opened 2025-01-08
VB-4197 Ensure potentially unsafe data is escaped before rendering

hutcheonb-moj:
* Remove unnecessary use of Nunjuck's `safe` filter
* Use `escape` filter where appropriate for potentially unsafe data
* Rework visitor address handling to no longer need `<br>` tags between lines. Now uses newlines.


server/views/pages/timetable.njk
```diff
@@ -29,11 +29,11 @@
       </ul>
     {% else %}
       {% if prisonerIncentiveLevelGroupNames | length %}
-        {{ prisonerIncentiveLevelGroupNames | join("<br>") | safe -}}
+        {{ prisonerIncentiveLevelGroupNames | join("<br>") -}}
```
tpmcgowan, line 32:
```suggestion
        {{ prisonerIncentiveLevelGroupNames | join("\n") | escape | nl2br | safe -}}
```
as discussed!

server/views/pages/visit/summary.njk
```diff
@@ -222,31 +222,31 @@
                   text: 'Name'
                 },
                 value: {
-                  html: '<span data-test="test-visitor-name' + loop.index + '">' + visitor.name | safe + ' (' + (visitor.relationshipDescription | lower) + ' of the prisoner)</span>'
+                  html: '<span data-test="visitor-name' + loop.index + '">' + visitor.name + ' (' + (visitor.relationshipDescription | lower) + ' of the prisoner)</span>'
```
tpmcgowan, line 225:
```suggestion
                  html: '<span data-test="visitor-name' + loop.index + '">' + visitor.name | escape + ' (' + (visitor.relationshipDescription | lower) + ' of the prisoner)</span>'
```

integration_tests/pages/visitDetails.ts
```diff
@@ -47,26 +47,26 @@ export default class VisitDetailsPage extends Page {
   visitEmail = (): PageElement => cy.get('[data-test="visit-email"]')
 
   // Visitor Details-1
-  visitorName1 = (): PageElement => cy.get('[data-test="test-visitor-name1"]')
+  visitorName1 = (): PageElement => cy.get('[data-test="visitor-name1"]')
```
tpmcgowan, line 50: All these updated names are better having removed the unnecessary `test-`. But when merged, worth running e2e test job in case it has selectors that need updating.

PR #916, opened 2025-01-17
VB-4976 Show open and closed visits on the same page

tpmcgowan:
* Show both open and closed sections for a given selected session template (unless capacity is zero)
* Handle `UNKNOWN` (labelled 'All') visits

### DIFF
--- a/server/views/pages/visitsByDate/visitsByDate.njk
+++ b/server/views/pages/visitsByDate/visitsByDate.njk
@@ -13,10 +13,86 @@
   {{ breadcrumbs() }}
 {% endblock %}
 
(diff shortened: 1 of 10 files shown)


server/views/pages/visitsByDate/visitsByDate.njk
```diff
@@ -13,10 +13,92 @@
   {{ breadcrumbs() }}
 {% endblock %}
 
+{% macro renderVisitsSection(type) %}
```
tpmcgowan, line 16: Because the page can now have multiple tables of visits (e.g. open/closed), moved the visit table code to this macro to avoid duplication.

PR #923, opened 2025-01-24
VB-4942 Filter visit notification events by configurable allow list

tpmcgowan:
Add config to filter visit review notifications to an allowed set of values. This means that additional events enabled in production, for example, won't necessarily come through on the front-end unless they are an allowed type.

### DIFF
--- a/server/config.ts
+++ b/server/config.ts
@@ -1,3 +1,5 @@
+import { NotificationType } from './data/orchestrationApiTypes'
+
 const production = process.env.NODE_ENV === 'production'
 
 function get<T>(name: string, fallback: T, options = { requireInProduction: false }): T | string {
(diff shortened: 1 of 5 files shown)


server/config.ts
```diff
@@ -130,6 +132,16 @@ export default {
       agent: new AgentConfig(Number(get('ORCHESTRATION_API_TIMEOUT_RESPONSE', 10000))),
     },
   },
+  features: {
+    notificationTypes: {
+      enabledNotifications: <NotificationType[]>(
+        get(
+          'FEATURE_ENABLED_NOTIFICATION_TYPES',
+          'PRISONER_RELEASED_EVENT,PRISONER_RECEIVED_EVENT,PRISON_VISITS_BLOCKED_FOR_DATE',
```
tpmcgowan, line 140: Just have this default config in for now, but could enable additional types via Helm config - e.g. for testing in dev environment.

PR #928, opened 2025-01-31
VB-5036 Refactor to split visit details page code

tpmcgowan:
Initial step to move the many varied routes in `visit.ts` into separate controllers.


hutcheonb-moj (approved):
LGTM

PR #929, opened 2025-01-31
VB-5151 Update table availability text on Select date/time page

tpmcgowan:
Change text for each slot from `X table(s) available` to `X table(s) available of Y table(s)`.


simonnebMOJ (approved):
LGTM

PR #935, opened 2025-02-06
VB-5150 Filter notification count

tpmcgowan:
Pass the enabled notification types when requesting notification count so that an accurate number is returned.

### DIFF
--- a/server/routes/testutils/testData.ts
+++ b/server/routes/testutils/testData.ts
@@ -186,17 +186,19 @@ export default class TestData {
   }: Partial<NotificationGroup> = {}): NotificationGroup => ({ reference, type, affectedVisits })
 
   static notificationVisitInfo = ({
-    bookedByName = 'User One',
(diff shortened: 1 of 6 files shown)


server/routes/testutils/testData.ts
```diff
@@ -186,17 +186,19 @@ export default class TestData {
...
   static notificationVisitInfo = ({
-    bookedByName = 'User One',
-    bookedByUserName = 'user1',
-    bookingReference = 'ab-cd-ef-gh',
     prisonerNumber = 'A1234BC',
+    bookedByUserName = 'user1',
     visitDate = '2023-11-01',
+    bookingReference = 'ab-cd-ef-gh',
+    bookedByName = 'User One',
+    notificationEventAttributes = [],
```
tpmcgowan, line 194: Unrelated - just tidying up some missing properties that were highlighted when type definition was updated.

hutcheonb-moj (approved):

PR #938, opened 2025-02-07
VB-5149 Filter notification types allowed on visit history

tpmcgowan:
Filter the items returned from the API for a visit's events audit to only include notification event types that are enabled.


hutcheonb-moj (approved):
LGTM

PR #948, opened 2025-02-21
VB-5218, refactor clear notifications

hutcheonb-moj:
Move clear notifications out of the visit.ts route, and into its own file


tpmcgowan (approved):
LGTM

PR #953, opened 2025-03-03
VB-5110 Visit notification warning summary component

tpmcgowan:
* Move visit notification messages to use warning component (based on error summary component but using orange)
* Move cancellation messages to above page heading


hutcheonb-moj (approved):
LGTM

PR #961, opened 2025-03-13
VB-4546, display MojAlert banner for 422 validation errors

hutcheonb-moj:
Redirect users to either confirm overbooking or select date and time page, and display new error message, if failing to book / update a visit

Includes VB-5147 - https://dsdmoj.atlassian.net/browse/VB-5147


server/views/pages/bookAVisit/dateAndTime.njk
```diff
@@ -54,6 +55,16 @@
     <div class="govuk-grid-column-full">
       {% include "partials/errorSummary.njk" %}
 
+      {% if validationMessage !== '' %}
+        {{ mojAlert({
+          variant: "error",
+          title: validationMessage,
+          showTitleAsHeading: true,
+          dismissible: false,
+          html: '<p data-test="validation-text">Select a new visit time.</p>'
```
tpmcgowan, line 64: Doesn't look like this `data-test` is actually used in any tests? (Or in the version of this in `overbooking.njk`) Looking at the documentation, it should be possible to set `attributes: { "data-test": "validation-error" } on the component. That way you can use the `text` instead of the `html` property too so you don't need add the HTML markup.

server/routes/visitJourney/dateAndTime.ts
```diff
@@ -102,10 +104,20 @@ export default class DateAndTime {
 
     visitSessionData.allowOverBooking = false // intentionally reset when returning to date and time page
 
+    const { validationError } = visitSessionData
```
tpmcgowan, line 107: The validation error is read from the session here, but does it also need to be cleared? E.g. what happens if I choose another session but the click the 'Back' link? Will I see the message still?

An alternative would be use `req.flash()` to set either just the validation error code or an object with the data needed to render the component (would need to potentially tweak `server/@types/express/index.d.ts` to allow this)

server/routes/visitJourney/dateAndTime.ts
```diff
@@ -102,10 +104,20 @@ export default class DateAndTime {
...
+
+    let validationMessage = ''
+    if (validationError === 'APPLICATION_INVALID_NON_ASSOCIATION_VISITS') {
+      validationMessage = `${prisonerName} now has a non-association on ${format(visitSessionData.visitSlot.startTimestamp, 'd MMMM')}.`
+    } else if (validationError === 'APPLICATION_INVALID_VISIT_ALREADY_BOOKED') {
+      validationMessage = `${prisonerName} now has another visit at ${format(parseISO(visitSessionData.visitSlot.startTimestamp), 'h:mmaaa').replace(':00', '')} on ${format(visitSessionData.visitSlot.startTimestamp, 'd MMMM')}.`
+    }
+
```
tpmcgowan, line 115: This kind of thing could potentially go in a utility function so determining message value for various conditions can be tested a lot more simply than at this level. Fine for now though as there are (currently!) only two options - and we'll be completely rewriting this page soon to do calendar view...

You might be able to simplify this code though with some of the functions in https://github.com/ministryofjustice/book-a-prison-visit-staff-ui/blob/3c829512d72d7b46152f52289b82aca754ce378a/server/utils/utils.ts. E.g. `prisonerTimePretty()` (badly named...)

server/routes/visitJourney/dateAndTime.ts
tpmcgowan: Thinking about it more, it feels like using `flash` is going to be a better option for this because that's exactly what it's for: setting a message that should be displayed once on a page to which the user is redirected. Otherwise you'll have to effectively re-implement what it does.

Need to think about a suitable key (`message`, `validationError` ?). If more than one gets set I think they come back as an array. When setting, could either:
* stick with setting just the error code (e.g. `req.flash('message', 'ERROR_CODE')`) which means the logic for building the actual message and testing this is dispersed

or


* set all the info at the point the error occurs. This means keeping related logic together and consolidated testing; and on pages that display messages you just read flash and render. E.g.
  * define type for MoJ Alert (e.g. `type MoJAlert = { variant: 'information' | 'error'; title: string; ... }`
  * set full details when error is caught: e.g. `const mojAlert: MoJAlert = { ... }; req.fash('message', mojAlert)`
  * (a utility function to take error code and `visitSessionData` and return an `MoJAlert` could abstract and simplify testing)
  * on pages to show the alerts, just read flash and then in template, for example, `{{ mojAlert(message) }}`

server/views/pages/bookAVisit/dateAndTime.njk
hutcheonb-moj: This data-test was used for initial implementation then I realised as the information given for that field is the same on a particular page it made no sense being so specific, I've reduced back to the `text` option

server/views/pages/bookAVisit/dateAndTime.njk
```diff
@@ -54,6 +55,10 @@
     <div class="govuk-grid-column-full">
       {% include "partials/errorSummary.njk" %}
 
+      {% if validationAlert | length %}
+        {{ mojAlert(validationAlert[0]) }}
+      {% endif %}
+
```
tpmcgowan, line 61: This is rendering full-width. Designs look to be two-thirds? (Applies to `overbooking.njk` too.) Worth double-checking what it should be.

server/routes/visitJourney/visitJourneyUtils.test.ts
```diff
@@ -8,3 +9,56 @@ describe('getUrlPrefix', () => {
...
+    const validationErrors: ApplicationValidationErrorResponse['validationErrors'] = [
+      'APPLICATION_INVALID_NON_ASSOCIATION_VISITS',
+    ]
+
+    const { mojAlert, url } = validationErrorsMojAlert(prisonerName, visitStartTimestamp, validationErrors)
+
+    expect(mojAlert).toStrictEqual({
+      text: 'Select a new visit time.',
+      showTitleAsHeading: true,
+      title: 'Smith, John now has a non-association on 12 March.',
```
tpmcgowan, line 27: Design shows prisoner name as `FIRSTNAME LASTNAME`. It also shows it in this format below the `H1` too.

We have another ticket coming up I believe about prisoner name format. So, worth checking if we should stick with what you have for now and change it all later; or, change it here in this PR. Could be fiddly because we just store `prisoner.name` in the session 🤔

server/views/pages/bookAVisit/dateAndTime.test.ts
```diff
@@ -210,6 +210,23 @@ describe('Views - Date and time of visit', () => {
...
 
+  it('should display 422 validation error - MojAlert', () => {
+    viewContext = {
+      validationAlert: [
+        {
+          text: 'Select whether to book for this time or choose a new visit time.',
+          showTitleAsHeading: true,
+          title: 'Another person has booked the last table.',
+          variant: 'warning',
+        },
```
tpmcgowan, line 221: Could use `TestData.mojAlert()` here.

server/routes/visitJourney/checkYourBooking.test.ts
```diff
@@ -247,39 +247,45 @@ testJourneys.forEach(journey => {
...
+                })
 
-              expect(visitSessionData.visitStatus).not.toBe('BOOKED')
-              expect(visitSessionData.visitReference).toBe(journey.isUpdate ? 'ab-cd-ef-gh' : undefined)
-              expect(auditService.bookedVisit).not.toHaveBeenCalled()
-            })
+                expect(visitSessionData.visitStatus).not.toBe('BOOKED')
+                expect(visitSessionData.visitReference).toBe(journey.isUpdate ? 'ab-cd-ef-gh' : undefined)
+                expect(auditService.bookedVisit).not.toHaveBeenCalled()
+              })
```
tpmcgowan, line 287: This test above has ended up nested within the wrong `describe()` block. It should be in the containing `Handle API errors` block as it isn't a 422 error (which has its own block)

server/@types/bapv.d.ts
```diff
@@ -122,6 +122,18 @@ export type VisitSessionData = {
   requestMethod?: ApplicationMethodType
   daysUntilBanExpiry?: number
   overrideBookingWindow?: boolean
+  validationError?: {
+    title: string
+    message: string
+    type: string
+  }
```
tpmcgowan, line 129: I don't think this is used?

integration_tests/mockApis/orchestration.ts
```diff
@@ -53,6 +54,31 @@ export default {
       },
     })
   },
+  stubBookVisitValidationFailed: ({
+    visit,
+    validationErrors = ['APPLICATION_INVALID_NON_ASSOCIATION_VISITS'],
+  }: {
+    visit: Visit
```
tpmcgowan, line 61: Don't actually need a `Visit` here; just `applicationReference: string` would do. This is because you don't return a `Visit` when the booking **fails** (unlike when a booking succeeds). 

Simplifying this means you don't need to set up and pass so much data when calling this stub in `checkYourBooking.cy.ts`.

integration_tests/e2e/checkYourBooking.cy.ts
```diff
@@ -268,4 +268,127 @@ context('Check visit details page', () => {
...
+      visit: TestData.visit({
+        startTimestamp: visitSessions[1].startTimestamp,
+        endTimestamp: visitSessions[1].endTimestamp,
+        visitContact: { name: 'Jeanette Smith', telephone: '09876 543 321' },
+        visitors: [
+          { nomisPersonId: contacts[0].personId, visitContact: true },
+          { nomisPersonId: contacts[1].personId, visitContact: false },
+        ],
+        visitorSupport: { description: 'Wheelchair ramp' },
+      }),
```
tpmcgowan, line 384: Could just pass an `applicationReference` here - see comment in `orchestration.ts`.

tpmcgowan (commented):
Looks good! Have left a few questions/suggestions.

Also need to update title and description of the PR.

server/routes/visitJourney/visitJourneyUtils.test.ts
hutcheonb-moj: we receive prisoner.name at the start of the journey (prisoner profile page) in that layout, think it's a little out of scope to re-build the name differently on 2 separate pages to satisfy the design

https://dsdmoj.atlassian.net/browse/VB-5147 - will be addressed as part of this

dhirajshettymoj (approved):
LGTM

PR #963, opened 2025-03-14
VB-5038 Filter visit details notifications

tpmcgowan:
Ensure that visit notifications (as well as visit events audit) are filtered by the enabled types.


hutcheonb-moj (approved):
LGTM

PR #968, opened 2025-03-19
VB-5042 Add custom heading spacing

tpmcgowan:
* Replace custom styles with those in use in the public service
* Remove some unnecessary spacing overrides
* Fix an issue on visit details page when grid columns get too narrow and content overlaps


hutcheonb-moj (approved):
LGTM

PR #971, opened 2025-03-20
VB-5299 Add room name to session schedule

tpmcgowan:
Show a session's visit room name below the visit time slot on the 'Select date and time' page.


hutcheonb-moj (approved):
lgtm

PR #973, opened 2025-03-21
VB-4411 Sort alerts on prisoner profile page by date

tpmcgowan:
* Use same date sort order for prisoner alerts on profile page as is used on visit details page (expiring soonest first)
* Remove unnecessary filter for `active | inactive` alerts (API does this)
* Rename `activeAlerts` to `alerts` for simplicity and consistency with restrictions

### DIFF
--- a/server/views/pages/prisoner/profile.njk
+++ b/server/views/pages/prisoner/profile.njk
@@ -63,7 +63,7 @@
         </li>
         <li>
           <strong>Alerts</strong>
-          <span data-test="active-alert-count">{{ activeAlertCount }} active</span>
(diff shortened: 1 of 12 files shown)


server/views/pages/prisoner/profile.njk
```diff
@@ -63,7 +63,7 @@
         </li>
         <li>
           <strong>Alerts</strong>
-          <span data-test="active-alert-count">{{ activeAlertCount }} active</span>
+          <span data-test="active-alert-count">{{ alerts | length }} active</span>
```
tpmcgowan, line 66: Have avoided passing an 'alert count' around everywhere by just using `| length` here

PR #975, opened 2025-03-24
VB-5112 Refactor update journey start

tpmcgowan:
* use new visit details endpoint to pre-populate session for update journey
* remove old 'visit history details' endpoint
* tidy and tweak session data passed through booking journey

### DIFF
--- a/server/@types/bapv.d.ts
+++ b/server/@types/bapv.d.ts
@@ -111,6 +111,7 @@ export type VisitSessionData = {
   visitorSupport?: VisitorSupport
   mainContact?: {
     contactId?: number
+    relationshipDescription?: string
     phoneNumber?: string
     email?: string
     contactName?: string
--- a/server/@types/bapv.d.ts
+++ b/server/@types/bapv.d.ts
@@ -109,7 +109,7 @@ export type VisitSessionData = {
   visitors?: VisitorListItem[]
   visitorSupport?: VisitorSupport
   mainContact?: {
-    contact?: VisitorListItem
+    contactId?: number
     phoneNumber?: string
     email?: string
     contactName?: string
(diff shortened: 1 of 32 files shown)


server/@types/bapv.d.ts
```diff
@@ -106,10 +106,12 @@ export type VisitSessionData = {
   visitSlot?: VisitSlot
   originalVisitSlot?: VisitSlot
   visitRestriction?: 'OPEN' | 'CLOSED'
+  visitorIds?: number[]
```
tpmcgowan, line 109: Add `visitorIds` to session so don't need to rely on having `visitors` populated. This means that at the start of an update journey, just the IDs can be used to pre-select existing visitors.

server/@types/bapv.d.ts
```diff
@@ -106,10 +106,12 @@ export type VisitSessionData = {
...
   visitors?: VisitorListItem[]
   visitorSupport?: VisitorSupport
   mainContact?: {
-    contact?: VisitorListItem
+    contactId?: number
+    relationshipDescription?: string
```
tpmcgowan, line 114: Instead of storing a `VisitorListItem`, just store the contact ID and relationship type which is all that is needed.

server/routes/visit/visitUtils.ts
```diff
@@ -1,22 +1,14 @@
...
-  locationDescription,
-}: {
-  prisonId: string
-  prisonName: string
-  cellLocation: string
-  locationDescription: string
-}) => {
-  if (prisonId === 'OUT') {
-    return locationDescription
+export const getPrisonerLocation = (prisoner: VisitBookingDetailsDto['prisoner']) => {
```
tpmcgowan, line 4: Just refactored this to take a whole 'prisoner' rather than specifying so many parameters.

hutcheonb-moj (approved):
LGTM

PR #976, opened 2025-03-25
VB-5293 Fix prisoner alert descriptions

tpmcgowan:
For prisoner alerts on Visit details page, use `alertCodeDescription` instead of `alertTypeDescription` in order to match what is displayed on Select visitor page.


hutcheonb-moj (approved):
LGTM

PR #977, opened 2025-03-27
VB-5364 Fix prisoner alerts/restrictions on update journey

tpmcgowan:
The way the feature flag to hide prisoner alerts/restrictions on visit details page was implemented prevented these being passed through to the update journey.

Change to just hide the alerts/restrictions in the visit details page template, based on the feature flag.


hutcheonb-moj (approved):
LGTM

PR #979, opened 2025-04-01
VB-5147, change how prisoner name is displayed

hutcheonb-moj:


server/routes/testutils/testData.ts
```diff
@@ -489,8 +489,8 @@ export default class TestData {
     },
     prisoner = {
       prisonerNumber: 'A1234BC',
-      firstName: 'JOHN',
-      lastName: 'SMITH',
+      firstName: 'John',
+      lastName: 'Smith',
```
tpmcgowan, line 493:
```suggestion
      firstName: 'JOHN',
      lastName: 'SMITH',
```

I think this needs to stay capitalised because this is how this endpoint returns this data (for the **prisoner** - visitor names are lower case). If we don't have it in capitals here we could potentially not lower case it and the tests would still pass.

server/routes/visit.ts
```diff
@@ -85,7 +84,8 @@ export default function routes({
     const visitSessionData: VisitSessionData = {
       allowOverBooking: false,
       prisoner: {
-        name: properCaseFullName(`${prisoner.lastName}, ${prisoner.firstName}`),
+        firstName: prisoner.firstName,
+        lastName: prisoner.lastName,
```
tpmcgowan, line 88:
```suggestion
        firstName: convertToTitleCase(prisoner.firstName),
        lastName: convertToTitleCase(prisoner.lastName),
```
The API returns prisoner name as **upper case** so it needs to be formatted.

server/routes/visitJourney/checkYourBooking.ts
```diff
@@ -18,6 +18,7 @@ export default class CheckYourBooking {
     const isUpdate = this.mode === 'update'
     const { visitSessionData } = req.session
     const { offenderNo } = visitSessionData.prisoner
+    const prisonerName = `${visitSessionData.prisoner.firstName} ${visitSessionData.prisoner.lastName}`
```
tpmcgowan, line 21: In the routes changed in this PR it looks like there are two different approaches being used to get prisoner name to a template:
1. build an (additional) `prisonerName` in the route and pass this to template to render as `{{ prisonerName }}`
2. pass, say, `prisoner` to the template and then render as `{{ prisoner.firstName + " " + prisoner.lastName }}`

Either is fine but it's probably worth being consistent. My slight preference would be that if the data is going to the template anyway then keep the route simpler by not creating an extra variable - and this keeps the presentational logic in the view.  Not much in it though: consistency is the main thing.

tpmcgowan (approved):
LGTM!

PR #980, opened 2025-04-03
VB-5172, change book endpoint on update journey

hutcheonb-moj:
On Update journey, use the new 'update' endpoint, no change to regular booking journey


server/data/orchestrationApiClient.ts
```diff
@@ -67,6 +67,23 @@ export default class OrchestrationApiClient {
     })
   }
 
+  async updateVisit(
+    applicationReference: string,
+    applicationMethod: ApplicationMethodType,
```
psoleckimoj, line 72: Couldn't you call this `applicationMethodType` since that's what it's being assigned to on line 79?

server/routes/visitJourney/checkYourBooking.test.ts
```diff
@@ -184,12 +185,23 @@ testJourneys.forEach(journey => {
           .expect(302)
           .expect('location', `${journey.urlPrefix}/confirmation`)
           .expect(() => {
-            expect(visitService.bookVisit).toHaveBeenCalledWith({
-              username: 'user1',
-              applicationReference: visitSessionData.applicationReference,
-              applicationMethod: visitSessionData.requestMethod,
-              allowOverBooking: false,
-            })
+            if (journey.isUpdate) {
```
psoleckimoj, line 188: Hmm, I think it would be better as a separate test as you're adding logic inside the test itself, even if it's a complete duplicate of this test. Otherwise we don't know what this test is doing just looking at the results. Not sure what @tpmcgowan thinks.

server/routes/visitJourney/checkYourBooking.test.ts
```diff
@@ -268,17 +282,27 @@ testJourneys.forEach(journey => {
...
-                applicationReference: visitSessionData.applicationReference,
-                applicationMethod: visitSessionData.requestMethod,
-                allowOverBooking: false,
-              })
-
               expect(visitSessionData.visitStatus).not.toBe('BOOKED')
               expect(visitSessionData.visitReference).toBe(journey.isUpdate ? 'ab-cd-ef-gh' : undefined)
               expect(auditService.bookedVisit).not.toHaveBeenCalled()
+
+              if (journey.isUpdate) {
```
psoleckimoj, line 289: Same as earlier.

server/services/visitService.ts
```diff
@@ -59,6 +59,29 @@ export default class VisitService {
...
+    applicationReference: string
+    applicationMethod: ApplicationMethodType
+    allowOverBooking: boolean
+  }): Promise<Visit> {
+    const token = await this.hmppsAuthClient.getSystemClientToken(username)
+    const orchestrationApiClient = this.orchestrationApiClientFactory(token)
+
+    const visit = await orchestrationApiClient.updateVisit(
+      applicationReference,
+      applicationMethod,
```
psoleckimoj, line 78: Re my earlier comment, I see why it has a different name now.

psoleckimoj: Also, the call to this service passes an object but when you call the client you're passing individual arguments? For consistency, I'd pass an object here but not sure how the rest of the code looks.

psoleckimoj (approved):
LGTM. Some comments.

server/services/visitService.ts
hutcheonb-moj: This is essentially a duplicate of the book visit endpoint, I think keeping them structured and variables named the same is beneficial in this case as it's very clear when looking at them side by side that they have same input / output

server/routes/visitJourney/checkYourBooking.test.ts
hutcheonb-moj: Will address this now, I'll admit it didn't feel correct at the time. Will pull this logic out into seperate test

psoleckimoj (dismissed):
LGTM!

tpmcgowan (approved):
LGTM

PR #999, opened 2025-05-02
VB-5426 Update for Alerts changes

tpmcgowan:
Names of date fields for `Alerts` will be changing in the orchestration service API (for retrieving visit details and prisoner profile). This is to add `updatedDate` and to make `startDate` and `expiryDate` consistent with `Restrictions`.

This change:
* updates the code to use the new values
* in the data layer it also includes a temporary remapping function so it will work with the old or new names (to be removed once the API changes are deployed)


server/@types/orchestration-api.d.ts
tpmcgowan, line 1: Auto-generated from Swagger - no need to review

server/data/orchestrationApiClient.ts
```diff
@@ -352,4 +358,15 @@ export default class OrchestrationApiClient {
...
     return { visitContact, mainContactId }
   }
+
+  // TODO remove when property rename API change deployed
+  private remapAlertsDates(alerts: Alert[]): Alert[] {
+    return alerts.map((alert: AlertDtoRaw) => {
+      return {
+        ...alert,
+        startDate: alert.startDate ?? (alert.dateCreated || null),
+        expiryDate: alert.expiryDate ?? (alert.dateExpires || null),
```
tpmcgowan, line 368: Handle incoming old and new API values but return the new ones. E.g. prefer `expiryDate` but if not set use `dateExpires` - or failing that, `null`

server/data/orchestrationApiTypes.ts
```diff
@@ -27,6 +27,11 @@ export type SessionSchedule = components['schemas']['SessionScheduleDto']
 export type VisitSession = components['schemas']['VisitSessionDto']
 
 export type Alert = components['schemas']['AlertDto']
+// TODO remove when property rename API change deployed
+export interface AlertDtoRaw extends Alert {
```
tpmcgowan, line 31: Temporarily adding a type that has old and new date property names

server/utils/visitorUtils.test.ts
```diff
@@ -32,7 +32,12 @@ describe('visitorUtils', () => {
       ['No restrictions', [], { isBanned: false }],
       ['No BAN restrictions', [TestData.restriction({ restrictionType: 'CLOSED' })], { isBanned: false }],
       [
-        'Single ban with no end date',
+        'Single ban with no end date (null)',
+        [TestData.restriction({ restrictionType: 'BAN', expiryDate: null })],
+        { isBanned: true },
+      ],
+      [
+        'Single ban with no end date (undefined)',
```
tpmcgowan, line 40: Noticed that when we get data directly from the contact registry, visitor restrictions with no end date do not have `expiryDate` set (i.e. it is `undefined`). When we get the same data back from orchestration it would be `null`. Adding test for both scenarios so behaviour same wherever the data has come from.

server/utils/visitorUtils.ts
```diff
@@ -53,7 +53,7 @@ export const getBanStatus = (restrictions: Restriction[]): BanStatus => {
     return { isBanned: false }
   }
   // if there is a ban with no end date no further checks needed
-  if (banned.find(ban => ban.expiryDate === undefined)) {
+  if (banned.find(ban => !ban.expiryDate)) {
```
tpmcgowan, line 56: As mentioned above, depending on the source, `expiryDate` could be `null` or `undefined`

psoleckimoj (approved):
LGTM!

PR #1003, opened 2025-05-06
VB-5359 Get offender restrictions from prisoner profile endpoint

tpmcgowan:
* Remove call to get offender restrictions from Prison API and instead get them when retrieving prisoner profile from orchestration service
* Prison API code now redundant so all removed


hutcheonb-moj (dismissed):
LGTM

hutcheonb-moj (approved):
lgtm - the merged PR was approved separately

PR #1008, opened 2025-05-09
VB 5393 Switch from MoJ Banner to Alert component for blocking dates (and flash types refactor)

tpmcgowan:
* Tidy up and make more consistent types for `req.flash()`
* Replace MoJ Banner component with Alert component when blocking visit dates

### DIFF
--- a/server/@types/bapv.d.ts
+++ b/server/@types/bapv.d.ts
@@ -11,6 +11,10 @@ import {
   VisitSummary,
 } from '../data/orchestrationApiTypes'
 
+type TextOrHtml = { text: string; html?: never } | { text?: never; html: string }
+
+export type FlashFormValues = Record<string, unknown>
+
 export type PrisonerDetailsItem = {
   html?: string
   text?: string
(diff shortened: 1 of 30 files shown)


server/@types/bapv.d.ts
```diff
@@ -89,14 +93,6 @@ export type VisitSlotList = {
   [key: string]: VisitSlotsForDay[] // keyed on month value, e.g. 'February 2022'
 }
 
-export type FormError = {
```
tpmcgowan, line 92: Replaced where used with more accurate `FieldValidationError` from `express-validator`

server/routes/testutils/appSetup.ts
```diff
@@ -51,6 +52,14 @@ export const user: PrisonUser = {
   activeCaseLoadId: 'HEI',
 }
 
+export type FlashData = {
+  errors?: ValidationError[]
+  formValues?: FlashFormValues[]
+  messages?: MoJAlert[]
+  // FIXME below needed for cancel visit route; remove when this is updated to store in req.session
+  [key: string]: string[] | ValidationError[] | FlashFormValues[] | MoJAlert[] | undefined
```
tpmcgowan, line 60: This TypeScript fudge needed because `req.flash()` is used to pass data to cancellation confirmation page. Not ideal because page breaks if it is refreshed (because flash disappears once read). Better to use `req.session`. Will sort in a follow-on PR....

psoleckimoj (approved):
LGTM!

PR #1009, opened 2025-05-09
Improve cancellation confirmation page

tpmcgowan:
* Use `req.session` instead of `req.flash` for storing data so it won't fail if page refreshed
* Update links from `/` to `/back-to-start` (which incorporates clearing session between journeys)
* Handle visiting cancellation confirmation page when no appropriate session data set


psoleckimoj (approved):
LGTM!

PR #1017, opened 2025-05-21
VB-3758 API changes for restrictions (add restrictionId) and move to v2 contacts API

tpmcgowan:
* update imported API definitions - to add `restrictionId` to contact (visitor) restrictions
* Move to the `/v2` contact registry API

### DIFF
--- a/server/@types/orchestration-api.d.ts
+++ b/server/@types/orchestration-api.d.ts
@@ -1491,12 +1491,12 @@ export interface components {
       comment?: string
       /**
        * @description Primary Address
-       * @example Y
+       * @example false
        */
       primary: boolean
       /**
        * @description No Fixed Address
-       * @example N
+       * @example false
        */
       noFixedAddress: boolean
       /**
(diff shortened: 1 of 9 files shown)


server/@types/orchestration-api.d.ts
tpmcgowan, line 1: Auto imported from Swagger - as is `prisoner-contact-registry.d.ts` - so no need to review

server/data/prisonerContactRegistryApiClient.ts
```diff
@@ -14,14 +14,13 @@ export default class PrisonerContactRegistryApiClient {
     )
   }
 
-  async getPrisonerSocialContacts(approvedVisitorsOnly: boolean, offenderNo: string): Promise<Contact[]> {
+  async getPrisonersApprovedSocialContacts(offenderNo: string): Promise<Contact[]> {
```
tpmcgowan, line 17: Removing the option to get 'unapproved' contacts because this is no longer needed: staff service only requests contacts directly from contact registry during booking journey which means we **always** want `approved`.

adaviesMOJ (approved):

PR #1019, opened 2025-05-22
VB-5416, add character count to additional support page

hutcheonb-moj:
Change additional support page to use 'Character Count' component

Added function which stores in session if currently updated visit is originally from the public service

### DIFF
--- a/server/@types/bapv.d.ts
+++ b/server/@types/bapv.d.ts
@@ -122,6 +122,7 @@ export type VisitSessionData = {
   requestMethod?: ApplicationMethodType
   daysUntilBanExpiry?: number
   overrideBookingWindow?: boolean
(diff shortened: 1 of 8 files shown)


server/@types/bapv.d.ts
```diff
@@ -122,6 +122,7 @@ export type VisitSessionData = {
   requestMethod?: ApplicationMethodType
   daysUntilBanExpiry?: number
   overrideBookingWindow?: boolean
+  publicBooker?: boolean
```
tpmcgowan, line 125:
```suggestion
  publicBooker?: boolean // true if booking originally made in public service
```

tpmcgowan (approved):
LGTM!

PR #1020, opened 2025-05-22
VB-5245 Refactor visit history timeline

tpmcgowan:
Refactor utility function that builds MoJ Timeline component entries from a visit's event audit history. 

* Moved from `VisitService` to dedicated file `visitEventsTimelineBuilder.ts` where it is used in the `./routes/visit/` folder
* Tests moved but unchanged 
* The logic for building items is complex but it has been split up and simplified
* Improved type safety
* as new visit notification types are supported, these will 'just work' without changes to this code


server/routes/visit/visitEventsTimelineBuilder.test.ts
```diff
@@ -0,0 +1,303 @@
...
+        text: 'Cancelled due to illness',
+      },
+    ]
+
+    params.events = [
+      {
+        type: 'CANCELLED_VISIT',
+        applicationMethodType: 'WEBSITE',
+        actionedByFullName: 'User One',
+        userType: 'STAFF',
```
tpmcgowan, line 180: This was previously `userType: 'PUBLIC'` but I have changed it to `STAFF`. This is the only change made to the existing tests (other than moving them!)

@hutcheonb-moj I am assuming this was incorrect in the original test data because a free text cancellation reason is only captured when a visit is cancelled in the staff service. Does that sound right?

server/routes/visit/visitEventsTimelineBuilder.ts
```diff
@@ -0,0 +1,82 @@
...
+  events,
+  visitNotes,
+}: {
+  events: EventAudit[]
+  visitNotes: VisitBookingDetailsDto['visitNotes']
+}): MojTimelineItem[] => {
+  // Filter to only supported event types and reverse order for display
+  const filteredEvents = events.filter(event => event.type in eventAuditTypes).reverse()
+
+  return filteredEvents.map((event, index) => {
```
psoleckimoj, line 27: Can't you just drop the `filter().reverse()` in here then there's no need to assign a new var? So `events.filter().reverse().map(....` ?

psoleckimoj (approved):
LGTM!

server/routes/visit/visitEventsTimelineBuilder.ts
tpmcgowan: Good suggestion. I've done this and it looks nicer now 🙂

psoleckimoj (approved):
LGTM!

PR #1021, opened 2025-05-22
VB-5416, correction, add threshold

hutcheonb-moj:
previous missed 

```Uses conditional reveal with character count component: [Character count – GOV.UK Design System](https://design-system.service.gov.uk/components/character-count/)

200 max characters```

on ticket


tpmcgowan (approved):
LGTM

PR #1022, opened 2025-05-23
VB-3758 Process raw API event and notification types

tpmcgowan:
The API returns notifications (e.g. `PRISONER_RELEASED`) and audit events (these include notification types and others like `CANCELLED`).

Most of these events have a one-to-one mapping to labels and functionality on the front-end. Visitor restrictions though can be local or global (`PERSON_RESTRICTION_UPSERTED_EVENT`, `VISITOR_RESTRICTION_UPSERTED_EVENT`) but these effectively have to be labelled and treated the same - as a `Visitor restriction`.

This change makes a distinction for events/notifications between raw data/types returned from the API and data/types for use in the application. In the data layer, the two visitor restriction types are converted to a custom value: `VISITOR_RESTRICTION`.

This means that:
* API can remain unchanged - so the distinction is maintained for future use / other services 
* impact on the application is minimal: existing functionality to list, group, filter events/notifications doesn't need to change
* visitor restriction events/notifications can be supported (next PR!)

(N.B. this is all preparatory work for supporting visitor restrictions - no feature changes in this)

### DIFF
--- a/server/data/orchestrationApiClient.ts
+++ b/server/data/orchestrationApiClient.ts
@@ -7,33 +7,36 @@ import {
   CancelVisitOrchestrationDto,
   ChangeApplicationDto,
   CreateApplicationDto,
+  EventAuditType,
+  EventAuditTypeRaw,
+  ExcludeDateDto,
   IgnoreVisitNotificationsDto,
   IsExcludeDateDto,
   NotificationCount,
   NotificationGroup,
+  NotificationGroupRaw,
+  NotificationType,
+  NotificationTypeRaw,
   PageVisitDto,
   PrisonDto,
   PrisonerProfile,
-  ExcludeDateDto,
   SessionCapacity,
   SessionSchedule,
   Visit,
+  VisitBookingDetails,
+  VisitBookingDetailsRaw,
   VisitPreview,
   VisitRestriction,
   VisitSession,
-  EventAuditType,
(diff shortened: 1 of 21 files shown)


server/data/orchestrationApiClient.ts
tpmcgowan, line 1: Key changes in this PR are to the methods in this file

server/routes/testutils/testData.ts
```diff
@@ -199,15 +202,20 @@ export default class TestData {
...
+    affectedVisits = [this.notificationVisitInfo()],
   }: Partial<NotificationGroup> = {}): NotificationGroup => ({ reference, type, affectedVisits })
 
+  // raw data with types as returned from API
+  static notificationGroupRaw = ({
+    reference = this.notificationGroup().reference,
+    type = this.notificationGroup().type as NotificationTypeRaw,
+    affectedVisits = this.notificationGroup().affectedVisits,
+  }: Partial<NotificationGroupRaw> = {}): NotificationGroupRaw => ({ reference, type, affectedVisits })
+
```
tpmcgowan, line 218: Here (and with `visitBookingDetails` below), I'm making two similar test data methods to distinguish between the raw data - for use in API client and Wiremock - and processed data for use everywhere else.

psoleckimoj (approved):
LGTM!

PR #1028, opened 2025-06-11
VB-5569 Needs review page: group by visit

tpmcgowan:
Rewrite of 'Needs review' visit listing page:
*  use new API that groups by visit, with potentially multiple notifications per visit (previously, a single visit with two notifications would mean two rows)
* reduplicates notification types within a visit row so only one of each type is shown

(Some tidying to do in a follow-on PR to remove now-redundant code for handling the old API and data format.)

### DIFF
--- a/integration_tests/mockApis/orchestration.ts
+++ b/integration_tests/mockApis/orchestration.ts
@@ -15,6 +15,7 @@ import {
   SessionSchedule,
   Visit,
   VisitBookingDetailsRaw,
+  VisitNotificationsRaw,
   VisitPreview,
   VisitSession,
 } from '../../server/data/orchestrationApiTypes'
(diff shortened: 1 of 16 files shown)


integration_tests/mockApis/orchestration.ts
```diff
@@ -394,6 +395,28 @@ export default {
...
+      },
+      response: {
+        status: 200,
+        headers: { 'Content-Type': 'application/json;charset=UTF-8' },
+        jsonBody: visitNotifications,
+      },
+    })
+  },
+
+  // TODO remove: endpoint deprecated
```
tpmcgowan, line 419: Will be stripping out old code in follow-on PR to keep simpler.

psoleckimoj (approved):
LGTM!

PR #1041, opened 2025-06-26
Express v5 upgrade

tpmcgowan:


server/routes/search.test.ts
```diff
@@ -221,13 +221,13 @@ describe('Prisoner search page', () => {
 
     describe('POST /search/prisoner', () => {
       it('should redirect to search results page with search query when no search term entered', () => {
-        return request(app).post('/search/prisoner').expect(302).expect('location', '/search/prisoner/results')
+        return request(app).post('/search/prisoner').send({}).expect(302).expect('location', '/search/prisoner/results')
```
tpmcgowan, line 224: Quite a few tests failed on upgrade where they were testing no data being sent by just not calling `send()`. Fixed by adding `.send({})` which is more realistic as `req.body` comes through as `{}` on `POST` (but in other cases can be `undefined`).

server/views/layout.njk
```diff
@@ -29,14 +29,14 @@
 {% endblock %}
 
 {% block beforeContent %}
-  <div role="navigation">
-    {% if backLinkHref %}
+  {% if backLinkHref %}
```
tpmcgowan, line 32: Unrelated - just fixing an accessibility issue the same as identified in public service: if no back link you would end up with an empty `navigation` region which could be confusing.

server/routes/search.test.ts
psoleckimoj: I think `.send` with JSON data needs changing to `.json`

psoleckimoj (approved):
LGTM!

PR #1047, opened 2025-07-08
VB-5756 Tweak display of prisoner scheduled events

tpmcgowan:


server/@types/prison-register-api.d.ts
tpmcgowan, line 1: Unrelated - just deleting a stray file that should have been removed as part of #1004

simonnebMOJ (approved):
LGTM

PR #1048, opened 2025-07-10
VB-5761 Visit details page: fix for visitors with no date of birth

tpmcgowan:
* Show "Not entered" if a visitor has no date of birth set
* Fix double-escaping of main contact name on check details and booking confirmation page

### DIFF
--- a/server/views/pages/bookAVisit/checkYourBooking.njk
+++ b/server/views/pages/bookAVisit/checkYourBooking.njk
@@ -138,7 +138,7 @@
             text: "Main contact"
           },
           value: {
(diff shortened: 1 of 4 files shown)


server/views/pages/bookAVisit/checkYourBooking.njk
```diff
@@ -138,7 +138,7 @@
             text: "Main contact"
           },
           value: {
-            html: '<p><span class="test-main-contact-name">' + contactName | escape+ '</span><br><span class="test-main-contact-number">' +
+            html: '<p><span class="test-main-contact-name">' + contactName + '</span><br><span class="test-main-contact-number">' +
```
tpmcgowan, line 141: No need to `escape` here because it is auto-escaped when the variable is defined on line 32 above.

PR #1051, opened 2025-07-16
VB-5773 Add 'Request visits' tile with count

tpmcgowan:
* For prisons that are enabled for `PUBLIC` show new 'Requested visits' tile with count of outstanding visit requests.

* Add `FEATURE_VISIT_REQUEST` flag that controls display of the new tile (currently only enabled for dev & staging)


server/data/testutils/mockFeature.ts
```diff
@@ -0,0 +1,14 @@
+import config from '../../config'
+
+type FeatureNames = keyof typeof config.features
+
+// eslint-disable-next-line import/prefer-default-export
+export const setFeature = <Feature extends FeatureNames>(
```
tpmcgowan, line 6: New test helper function to override values in `config.features`. Typing with generics means only keys (and their typed values) set in `config.ts` will work - and when these change it will cause TypeScript errors in tests to highlight that they need updating.

server/routes/testutils/testData.ts
```diff
@@ -354,15 +355,16 @@ export default class TestData {
...
-    prisonId = 'HEI',
-    prisonName = 'Hewell (HMP)',
-    active = true,
-    policyNoticeDaysMax = 28,
-    policyNoticeDaysMin = 2,
-    maxTotalVisitors = 6,
-    maxAdultVisitors = 3,
-    maxChildVisitors = 4,
-    adultAgeYears = 18,
+    prisonId = this.prisonDto().code,
```
tpmcgowan, line 358: `Prison` type is effectively a clone of raw `PrisonDto` from API so keeping default test data values consistent between the two.

server/@types/orchestration-api.d.ts
tpmcgowan, line 1: Auto-generated from Swagger. No need to review.

hutcheonb-moj (approved):
LGTM, new way of doing feature flag for tests looks much more elegant than our previously used method

psoleckimoj (approved):
LGTM!

PR #1055, opened 2025-07-17
VB-5777 Requested visits listing page

tpmcgowan:
Add Requested visits listing page (wrapped in `FEATURE_VISIT_REQUEST` so only on dev & staging currently).

### DIFF
--- a/server/services/visitNotificationsService.ts
+++ b/server/services/visitNotificationsService.ts
@@ -4,7 +4,6 @@ import {
   NotificationCount,
   Visit,
   VisitNotifications,
-  VisitRequestsCountDto,
 } from '../data/orchestrationApiTypes'
 
 export default class VisitNotificationsService {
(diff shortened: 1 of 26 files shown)


server/services/visitNotificationsService.ts
```diff
@@ -20,13 +19,6 @@ export default class VisitNotificationsService {
     return orchestrationApiClient.getNotificationCount(prisonId)
   }
 
-  async getVisitRequestCount(username: string, prisonId: string): Promise<VisitRequestsCountDto> {
```
tpmcgowan, line 23: Moved this to new `VisitRequestService` as there will be more request-related stuff coming.

server/views/pages/request/visitRequestsListing.njk
```diff
@@ -0,0 +1,124 @@
...
+{% from "govuk/components/table/macro.njk" import govukTable %}
+
+{% set pageHeaderTitle = "Requested visits" %}
+
+{% block beforeContent %}
+  {{ breadcrumbs() }}
+{% endblock %}
+
+{% set requestListRows = [] %}
+{% for visitRequest in visitRequests %}
```
psoleckimoj, line 12: Just thinking, for clarity in the template, could this data not be created in the route then passed through?

tpmcgowan: Yeah, might be worth a try 🤔. Just built rows in the template out of habit tbh as we've tended to do that - but it's not particularly pretty. I am using Nunjucks filters in places to build it up - but our own ones at least are just calls to `utils.ts` etc anyway so would work. Would need to think about escaping as you need to do this manually when passing an `html` value to a GOV.UK macro...

I'll merge this as it is for now but will try refactoring as suggested as a separate PR to see what it looks like 👍

PR #1057, opened 2025-07-18
VB-5790 Confirm update prompt: align with minimum booking window

tpmcgowan:
Follow on from #1053.

* Align 'Confirm update' prompt page with minimum booking window
* (plus an unrelated table styling fix)

### DIFF
--- a/server/routes/visit/updateVisitController.test.ts
+++ b/server/routes/visit/updateVisitController.test.ts
@@ -26,7 +26,7 @@ jest.mock('../visitorUtils', () => {
 })
 
 beforeEach(() => {
-  const fakeDate = new Date('2022-01-01')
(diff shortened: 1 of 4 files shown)


server/routes/visit/updateVisitController.test.ts
```diff
@@ -26,7 +26,7 @@ jest.mock('../visitorUtils', () => {
 })
 
 beforeEach(() => {
-  const fakeDate = new Date('2022-01-01')
+  const fakeDate = new Date('2022-01-01T08:00:00')
```
tpmcgowan, line 29: Not specifying a time on the fake date can do odd things because it'll default to `00:00:00` and you get into timezone issues.

dhirajshettymoj (approved):
LGTM

PR #1061, opened 2025-07-22
VB-5809 Visit request details

tpmcgowan:
* Update title and show alert on visit details page if the visit is a **request**
* Tidy logic around getting and grouping alerts on visit details page


server/routes/visit/visitUtils.ts
```diff
@@ -129,6 +146,21 @@ export const getVisitNotificationsAlerts = (notifications: VisitBookingDetails['
   return alerts
 }
 
+export const getVisitAlerts = (visitDetails: VisitBookingDetails): MoJAlert[] => {
```
tpmcgowan, line 149: Just combining the output of the three functions above, which are no longer exported.

PR #1073, opened 2025-08-14
Feature/vb 5838 update visit status labels

psoleckimoj:


server/routes/prisoner/prisoner.test.ts
```diff
@@ -343,7 +371,6 @@ describe('/prisoner/:offenderNo - Prisoner profile', () => {
           expect($('label[for="vo-override"]').text()).toContain('The prisoner has no available visiting orders')
           expect($('[data-test="book-a-visit"]').length).toBe(1)
           expect(flashProvider).toHaveBeenCalledWith('errors')
-          expect(flashProvider).toHaveBeenCalledTimes(1)
```
psoleckimoj, line 346: Removed because messages is also used but is meaningless in this test and swapping it to 2 also makes the test meaningless. The fact it's been called with `errors` means it's been called with the relevant data anyways.

PR #1088, opened 2025-09-03
VB-5322 Add new visit sessions and schedule endpoint

tpmcgowan:
Add new endpoint that will support the visit calendar page.


hutcheonb-moj (approved):
lgtm

PR #1090, opened 2025-09-04
VB-5921 Visit booking details - tweak title and timeline for requests

tpmcgowan:
* Fix page title to use `request` or `booking` as appropriate 
* handle auto-rejected requests in the visit history timeline


server/constants/eventAudit.ts
```diff
@@ -7,6 +7,7 @@ const eventAuditTypes: Partial<Record<EventAuditType, string>> = {
   REQUESTED_VISIT: 'Requested',
   REQUESTED_VISIT_APPROVED: 'Approved',
   REQUESTED_VISIT_REJECTED: 'Rejected',
+  REQUESTED_VISIT_AUTO_REJECTED: 'Rejected',
```
tpmcgowan, line 10: This is the label on the visit history timeline. Assuming this should be `Rejected` and not something like `Auto-rejected`? Can't see any relevant design...

adaviesMOJ (approved):

PR #1100, opened 2025-09-19
VB-5964 Replace maintenance page

tpmcgowan:
Add maintenance page with service unavailable page (with optional end date and time).

Maintenance page is controlled by environment variables that can be set via Helm config or edited manually via `kubectl` (documented in the `README`).

This replaces the previous version of the maintenance page that relied on shutting down the application and running a dedicated maintenance page container.


server/app.ts
```diff
@@ -58,6 +59,9 @@ export default function createApp(services: Services): express.Application {
     }),
   )
   app.use(setUpCurrentUser())
+
+  app.use(maintenancePageRoute())
```
tpmcgowan, line 63: Adding it here specifically so that DPS components (header/footer) have loaded but before any other services are called. E.g. the next line - `populateSelectedEstablishment()` needs orchestration/visit scheduler to be working...

PR #1292, opened 2026-04-14
Upgrade  to TypeScript v6

tpmcgowan:


package.json
```diff
@@ -98,12 +98,11 @@
     "lint-staged": "^16.4.0",
     "nock": "^14.0.12",
     "nodemon": "^3.1.14",
-    "openapi-typescript": "^7.13.0",
```
tpmcgowan, line 101: Removed this as not compatible with TypeScript v6 - and is installed on-demand when needed by script that generates type definitions.

psoleckimoj (approved):
lgtm!

PR #1295, opened 2026-04-24
VB-6611, hide alerts and restrictions

hutcheonb-moj:


server/routes/testutils/testData.ts
```diff
@@ -713,6 +715,8 @@ export default class TestData {
     visitors = this.visitBookingDetails().visitors,
     events = this.visitBookingDetails().events as VisitBookingDetailsRaw['events'],
     notifications = this.visitBookingDetails().notifications as VisitBookingDetailsRaw['notifications'],
+    skipAlertsAndRestrictions = this.visitBookingDetails()
+      .skipAlertsAndRestrictions as VisitBookingDetailsRaw['skipAlertsAndRestrictions'],
```
tpmcgowan:
```suggestion
    skipAlertsAndRestrictions = this.visitBookingDetails().skipAlertsAndRestrictions,
```

Haven't tested but suspect you don't need the type cast here.

server/utils/nunjucksSetup.ts
```diff
@@ -18,6 +18,7 @@ export default function nunjucksSetup(app: express.Express, applicationInfo: App
   app.locals.asset_path = '/assets/'
   app.locals.applicationName = config.applicationName
   app.locals.environmentName = config.environmentName
+  app.locals.dpsContacts = config.dpsContacts
```
tpmcgowan: I don't think you're actually using this Nunjucks global?

assets/sass/local.scss
```diff
@@ -330,3 +330,16 @@ p.bapv-visit-details__restriction--flagged {
...
+    margin-bottom: 0;
+  }
+
+  margin-top: 0;
+  margin-bottom: 0;
+}
+
+.govuk-inset-text--prisoner {
+  margin-bottom: -25px;
+}
```
tpmcgowan: I assume that this makes it match the design in terms of bottom of the inset having basically no space before the following heading. I think we need to challenge that though: it is too close; needs a small bottom margin.

Anyway, avoid overriding `govuk-` classes like this because it will affect all instances (current or future). There is a lot of styling for visit details page from [here](https://github.com/ministryofjustice/book-a-prison-visit-staff-ui/blob/e33eba07eae030da3a17648f851b8346f22645d3/assets/sass/local.scss#L224-L297) onwards. You'll see there's a page level class to scope stuff.

So, I'd suggest adding styles at an appropriate place within that block. One option would be something like:
```
.bapv-visit-details .govuk-inset-text {
  // styles here now scoped to only affect the Inset component within visit details page
}
```

BUT, better is probably to give a specific semantic class to the prisoner/visitor inset block. E.g. `bapv-visit-details__prisoner-alerts-hidden` and `bapv-visit-details__visitor-alerts-hidden`. You then would just define styles for these at a relevant place in the code block referenced above.

server/routes/visit/visitDetailsController.ts
```diff
@@ -33,6 +39,14 @@ export default class VisitDetailsController {
         return res.render('pages/visit/visitDetailsWrongEstablishment', { prison, reference, selectedEstablishment })
       }
 
+      const hideAlertsInsetHtml = visitDetails.skipAlertsAndRestrictions
+        ? hideAlertsInset({
+            startTimestamp: visitDetails.startTimestamp,
+            visitPrisonId: visitDetails.prison.prisonId,
+            prisonerPrisonId: visitDetails.prisoner.prisonId,
+          })
+        : ''
```
tpmcgowan: Get `skipAlertsAndRestrictions` to optionally return `null`. Then you just pass the return value from your function to the template rather than sending two things as below.

In the template, either wrap the GOV.UK macro with an `{{ if (condition) }}` block like you have; or `{{ govukInsetText( ... ) if hideAlertsInsetHtml }}`.

server/routes/visit/visitUtils.ts
```diff
@@ -252,3 +252,22 @@ export const isPublicBooking = (events: EventAudit[]): boolean => {
...
+  prisonerPrisonId,
+}: {
+  startTimestamp: VisitBookingDetails['startTimestamp']
+  visitPrisonId: string
+  prisonerPrisonId: string
+}): string => {
+  const visitStartTime = new Date(startTimestamp)
+
+  if (isPast(visitStartTime))
+    return `<p>Alerts and restrictions are not shown for past visits.</p><p>You can view alerts and restrictions for past visits in the <a href="${config.dpsContacts}">contacts service</a>.`
```
tpmcgowan: Missing a closing `<p>`

tpmcgowan (changes requested):
I've left some minor comments but there's a problem with this implementation in that it is using the same text for Prisoner / Visitor - in Figma they are different!

server/routes/visit/visitUtils.test.ts
```diff
@@ -509,4 +511,94 @@ describe('Visit utils', () => {
       expect(publicBooker).toStrictEqual(false)
     })
   })
+
+  describe('getHideAlertsInset', () => {
+    it('should return null if alerts not to be hidden', () => {
+      const fakeDate = new Date('2026-01-01T10:00:00') // 01-01-26 2PM
+      jest.useFakeTimers({ advanceTimers: true, now: fakeDate })
```
tpmcgowan: These two lines are in each test so you could put in a `beforeEach()` in this `describe()` block - with the `jest.useRealTimers()` in an `afterEach()`

server/routes/visit/visitUtils.ts
```diff
@@ -252,3 +252,58 @@ export const isPublicBooking = (events: EventAudit[]): boolean => {
...
+        classes: 'inset-text-prisoner',
+      },
+      visitor: {
+        html: 'Visitor restrictions are not shown for released prisoners.',
+        attributes: { 'data-test': 'visitor-inset' },
+      },
+    }
+  }
+
+  if (prisonerPrisonId !== visitPrisonId) {
```
tpmcgowan: `prisonerPrisonId !== 'TRN' && prisonerPrisonId !== visitPrisonId` - to not show this if transfer is actually in progress

or use `inOutStatus` so typing is better

tpmcgowan (approved):
LGTM!

PR #1296, opened 2026-04-24
VB-6637 Serve favicon at /favicon.ico

tpmcgowan:
Safari (WebKit) may initiate browser‑level requests to /favicon.ico independently of page navigation and before HTML parsing. This request gets caught in the authentication redirect. Because /favicon.ico doesn’t actually exist (GOV.UK template specifies a different path), the user then ends up on an error page (404 - ‘Sorry there is a problem with the service’).

This change serves the GOV.UK favicon at `/favicon.ico` to fix this.


psoleckimoj (approved):
lgtm

PR #1308, opened 2026-05-07
VB-5252 / VB-5253 / flag alerts that are added / updated

hutcheonb-moj:
updated 'getIdsToFlag' to always be a string array
returns ''ALERT_UUID'' when given `PRISONER_ALERT_UPDATED_EVENT / PRISONER_ALERT_CREATED_EVENT`


server/constants/notifications.ts
```diff
@@ -12,6 +14,8 @@ export const notificationTypes: Partial<Record<NotificationType, string>> = {
 
 // used on visits to review listing page
 export const notificationTypeReasons: Partial<Record<NotificationType, string>> = {
+  PRISONER_ALERT_CREATED_EVENT: 'the prisoner has a new or updated alert that may impact visits',
+  PRISONER_ALERT_UPDATED_EVENT: 'the prisoner has a new or updated alert that may impact visits',
```
tpmcgowan: This will presumably show two identical bullet points in the details component? 

Comment out one of these lines so both constants are there for reference but we only get the joint message once.

server/routes/visit/visitUtils.ts
```diff
@@ -248,6 +271,25 @@ export const getIdsToFlag = ({
   return Array.from(flaggedIds)
 }
 
+export function getStringIdsToFlag({
```
tpmcgowan: I don't think we need to duplicate this function here. It's just more to maintain and they are both doing essentially the same thing.

The original function only returned `number[]` because the relevant IDs happened to be numerical (although `attributeValue` from the API itself is always `string`; hence the `parseInt()`). I don't think they need to be integers? We're not doing any counting/sorting; just rendering as HTML attribute values so getting re-cast back to strings anyway.

So suggestion is: just keep the original function; set return to `string[]` and remove the `parseInt()`.

(If there is a reason to support the two return types it'd be better to have `number[] | string[]` as the return and have a type-guard to determine whether to parse as integer or not.)

server/views/pages/visit/visitDetails.njk
```diff
@@ -208,25 +208,30 @@
           </p>
           <div class="bapv-visit-details__alerts">
             {% for alert in visitDetails.prisoner.prisonerAlerts %}
+              {% if alert.alertUuid in flaggedAlertCreatedIds %}
+                <p class="bapv-visit-details__alert--flagged" id="prisoner-alert={{ alert.alertUuid }}">This new alert has been added</p>
+              {% elseif alert.alertUuid in flaggedAlertUpdatedIds %}
+                <p class="bapv-visit-details__alert--flagged">This alert has been updated</p>
+              {% endif %}
```
tpmcgowan, line 216:
```suggestion
              {% if alert.alertUuid in flaggedAlertCreatedIds %}
                <p class="bapv-visit-details__alert--flagged" id="prisoner-alert-{{ alert.alertUuid }}">This new alert has been added</p>
              {% endif %}
              {% if alert.alertUuid in flaggedAlertUpdatedIds %}
                <p class="bapv-visit-details__alert--flagged">This alert has been updated</p>
              {% endif %}
```

Typo to fix and splitting to two if/else blocks as they're logically distinct the two sets of IDs.

tpmcgowan (approved):
LGTM

PR #1318, opened 2026-05-12
VB-6674 Handle deleted visitor relationships

tpmcgowan:
On the visit details page, if a visitor relationship comes through as `null` then display `Relationship deleted` instead of empty string.


adaviesMOJ (approved):
Nice!

PR #1324, opened 2026-05-20
VB-6682 Print visit passes: add APIs and button (feature flagged)

tpmcgowan:
* Add API support for getting visit pass data
* Add config for feature flag (**dev only** so far)
* Add feature-flagged button for printing passes (doesn't do anything yet!)


hutcheonb-moj (approved):
lgtm

PR #1335, opened 2026-06-01
Fix localhost and staging environment tags

tpmcgowan:
* Fix CSS that adds a `STAGING` label to the header in that environment. (DPS doesn't have staging so this was achieved via CSS content; but broke as attribute names changed in DPS header.)
* Add `LOCALHOST` as an environment label (if `ENVIRONMENT_NAME=localhost` is set in `.env`).

### DIFF
--- a/server/config.ts
+++ b/server/config.ts
@@ -12,14 +12,6 @@ function get<T>(name: string, fallback: T, options = { requireInProduction: fals
   throw new Error(`Missing env var ${name}`)
 }
 
-function translateEnvironment(gitHubEnvironmentName: string = ''): string {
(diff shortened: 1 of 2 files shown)


server/config.ts
```diff
@@ -12,14 +12,6 @@ function get<T>(name: string, fallback: T, options = { requireInProduction: fals
   throw new Error(`Missing env var ${name}`)
 }
 
-function translateEnvironment(gitHubEnvironmentName: string = ''): string {
```
tpmcgowan, line 15: Removing this function that upper-cased the environment name: simpler to just have it lower case in the CSS!

hutcheonb-moj (approved):
lgtm

PR #1345, opened 2026-06-10
VB-6415: Move to GovUK 6 and MoJFE 9

psoleckimoj:


assets/sass/components/_alerts-and-restrictions.scss
```diff
@@ -32,7 +32,7 @@
   }
 
   &--DIHCON {
-    @extend .govuk-tag--blue;
+    @extend .govuk-tag--blue !optional;
```
psoleckimoj: This came through as a warning saying that the tag didn't exist. Didn't have any issues wth the others 🤔

tpmcgowan: Blue is the default tag colour. It [used to](https://github.com/alphagov/govuk-frontend/blob/d6656ab6d2f81a1d51344f57e20e3287b7b862e6/packages/govuk-frontend/src/govuk/components/tag/_index.scss#L61-L64) still have it's own class anyway but I don't think it [does now](https://github.com/alphagov/govuk-frontend/blob/e168cdb7fce16a9cd7be6117a73bd134279e4f45/packages/govuk-frontend/src/govuk/components/tag/_mixin.scss#L44).

So I suspect you could either just put `.govuk-tag` instead of `...--blue`; or, just have no style rules and a comment to say `DHICON` uses default blue.

psoleckimoj: I think that's now fixed 🙂

tpmcgowan (approved):
LGTM!

PR #1348, opened 2026-06-12
VB-6682, VB-6689 Visit pass tweaks

tpmcgowan:
* Prevent printing of visit passes for dates in the past
* Don't show print button if a visit has outstanding notifications


psoleckimoj (approved):
lgtm

PR #1350, opened 2026-06-15
VB-6686 Log visit pass printing to HMPPS Audit

tpmcgowan:


server/services/auditService.test.ts
```diff
@@ -219,30 +219,6 @@ describe('Audit service', () => {
     })
   })
 
-  it('sends a visit list printed audit message', async () => {
```
tpmcgowan, line 222: Removing an old audit service method here that was implemented years ago but never used.

psoleckimoj (approved):
lgtm

PR #1362, opened 2026-06-22
VB-6270 Refactor visit details navigation

tpmcgowan:
The visit details page is linked to from many other pages: prisoner profile, visit requests, visit reviews, visits by date, etc. 

Maintaining navigation state between these (for back links and redirects) had grown over time and become quite messy and inconsistent.

This change implements a more secure, allow-list based navigation state management. It is focussed on key journeys to/from the visit details page.


psoleckimoj (approved):
lgtm

PR #1364, opened 2026-06-24
VB-6783 Fix telemetry to ensure AppTraces is populated

tpmcgowan:
Fix log data not appearing in `AppTraces` in Application Insights. Caused by Importing `logger` before `initialiseTelemetry()` was called.


server/utils/azureAppInsights.ts
```diff
@@ -12,14 +11,13 @@ initialiseTelemetry({
   .addModifier(telemetry.processors.enrichSpanNameWithHttpRoute())
   .startRecording()
 
-const shutdown = async (signal: string) => {
-  logger.info(`${signal} received, shutting down...`)
```
tpmcgowan, line 16: Fixing by just removing the log line; `flushTelemetry()` logs internally anyway so not really losing anything.

adaviesMOJ (approved):

PR #1365, opened 2026-06-26
VB-6733 - Set remand visit limit and renewal day

hutcheonb-moj:
Set limit for number of visits remand prisoners can have each week
Set the day which the remand limit resets


server/routes/visitAllowances/updateAllowancesController.ts
```diff
@@ -0,0 +1,82 @@
...
+        operationId: res.locals.appInsightsOperationId,
+      })
+
+      req.flash('messages', {
+        variant: 'success',
+        title: 'Visit allowances updated',
+        text: `${remandVisitLimitPerWeek} visits every 7 days. This allowance renews on ${properCase(weekStartDay)}.`,
+      })
+      return res.redirect('/visit-allowances')
+    }
```
tpmcgowan, line 70: I think a success alert is a good idea but because it isn't actually on the ticket I've asked Ed (on Jira) to confirm.

server/views/pages/visitAllowances/update.njk
```diff
@@ -0,0 +1,107 @@
+{% extends "layout.njk" %}
+{% from "partials/breadcrumbs.njk" import breadcrumbs with context %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+{% from "govuk/components/input/macro.njk" import govukInput %}
+{% from "govuk/components/radios/macro.njk" import govukRadios %}
+
+{% set pageHeaderTitle = "Visit allowances for unconvicted prisoners" %}
```
tpmcgowan:
```suggestion
{% set pageHeaderTitle = "Visit allowance for unconvicted prisoners" %}
```

server/views/pages/visitAllowances/update.njk
```diff
@@ -0,0 +1,107 @@
...
+{% block content %}
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-full">
+
+      {% include "partials/errorSummary.njk" %}
+
+      <h1 class="govuk-heading-l">{{ pageHeaderTitle }}</h1>
+
+      <h2 class="govuk-heading-m">Unconvicted prisoners</h2>
+
```
tpmcgowan:
```suggestion
```

This heading isn't on the design

server/views/pages/visitAllowances/update.njk
```diff
@@ -0,0 +1,107 @@
...
+          autocomplete: "off",
+          value: formValues.remandVisitLimitPerWeek,
+          errorMessage: errors | findError('remandVisitLimitPerWeek')
+        }) }}
+
+        {{ govukRadios({
+          name: "weekStartDay",
+          fieldset: {
+            legend: {
+              text: "On which day does the allowances renew?",
```
tpmcgowan:
```suggestion
              text: "On which day does the allowance renew?",
```

Typo!

server/constants/bapvUserRoles.ts
```diff
@@ -1,6 +1,8 @@
 const bapvUserRoles = {
   STAFF_USER: 'MANAGE_PRISON_VISITS',
   BOOKER_ADMIN: 'VISITS_BOOKER_ADMIN',
+  PRISON_IEP_ADMIN: 'MAINTAIN_PRISON_IEP_LEVELS',
+  GLOBAL_IEP_ADMIN: 'MAINTAIN_INCENTIVE_LEVELS',
```
tpmcgowan: Only one of these new roles is used so probably only need to add that one?

(Longer term we should be moving to the HMPPS permissions library so this kind of role handling will be replaced...)

server/routes/visitAllowances/viewAllowancesController.ts
```diff
@@ -1,19 +1,31 @@
...
     return async (req, res) => {
       const { prisonId } = req.session.selectedEstablishment
+      const { username, userRoles } = res.locals.user
       const incentiveLevels = await this.visitAllowanceService.getPrisonIncentiveLevels({
-        username: res.locals.user.username,
+        username,
         prisonId,
       })
 
+      const prisonConfig = await this.visitAllowanceService.getRemandConfig({
```
tpmcgowan: This and the call above are independent so you can parallelise them with `Promise.all()`

server/routes/visitAllowances/viewAllowancesController.ts
```diff
@@ -1,19 +1,31 @@
...
+        username,
+        prisonId,
+      })
+
+      const iepManagementRole = userRoles.includes(bapvUserRoles.PRISON_IEP_ADMIN)
```
tpmcgowan: Might be clearer if names `userHasIepManagementRole` or similar so it's clear that it is something about the user rather than just the role itself

server/views/pages/visitAllowances/view.njk
```diff
@@ -72,11 +74,16 @@
         ],
         rows: displayRows
       }) }}
+      {% if iepManagementRole %}
+        <p><a data-test="incentives-service-url" class="govuk-link--no-visited-state" href="{{ dpsIncentives }}prison-incentive-levels">
+          Go to the incentives service to change visit allowances for convicted prisoners.
+        </a></p>
+      {% else %}
+          <p><a data-test="incentives-service-url" class="govuk-link--no-visited-state" href="{{ dpsIncentives }}">
+          Visit allowances for convicted prisoners can be changed in the incentives service, but requires an extra user role.
```
tpmcgowan, line 83: Has this text actually come from Simon or Ed?

Not sure it's good UX to link them to a service where they won't be able to complete their task. My reading of https://dsdmoj.atlassian.net/browse/VB-6319?focusedCommentId=790638 is that if they don't have the role they just get a message; not a link. Might be wrong though!

server/views/pages/visitAllowances/view.njk
```diff
@@ -72,11 +74,16 @@
         ],
         rows: displayRows
       }) }}
+      {% if iepManagementRole %}
+        <p><a data-test="incentives-service-url" class="govuk-link--no-visited-state" href="{{ dpsIncentives }}prison-incentive-levels">
```
tpmcgowan, line 78: No test for this URL although there is a `data-test` attribute. In the route tests you can override `userSupplier()` to mock different user roles. An example is [here](https://github.com/ministryofjustice/book-a-prison-visit-staff-ui/blob/3b1a94da86e7347eb0f2352d3fb78e491789e428/server/routes/bookerManagement/booker/linkVisitorController.test.ts#L38).

server/routes/visitAllowances/updateAllowancesController.ts
```diff
@@ -0,0 +1,83 @@
...
+        errors: req.flash('errors'),
+      })
+    }
+  }
+
+  public change(): RequestHandler {
+    return async (req, res) => {
+      const { username } = res.locals.user
+      const { prisonId } = req.session.selectedEstablishment
+      const { weekStartDay, remandVisitLimitPerWeek } = req.body
```
tpmcgowan: Reading the form values off `req.body` here **before** checking for validation errors. With `express-validator` the pattern we typically use is to check for errors then get the form responses via `matchedData()`.

So, after validation if you do:
```
const data = matchedData<PrisonRemandConfig>(req)
```
then data will be a `PrisonRemandConfig` object. Or, even simpler:
```
const { weekStartDay, remandVisitLimitPerWeek } = matchedData<PrisonRemandConfig>(req)
```

server/routes/visitAllowances/updateAllowancesController.ts
```diff
@@ -0,0 +1,83 @@
...
+        text: `You changed the visit allowance for unconvicted prisoners.`,
+      })
+      return res.redirect('/visit-allowances')
+    }
+  }
+
+  public validate(): ValidationChain[] {
+    return [
+      body('weekStartDay')
+        .isIn(['MONDAY', 'TUESDAY', 'WEDNESDAY', 'THURSDAY', 'FRIDAY', 'SATURDAY', 'SUNDAY'])
```
tpmcgowan: In the admin service we define the days of the week as a constant (see [here](https://github.com/ministryofjustice/hmpps-visits-internal-admin-ui/blob/3b1606699482c8897a6e9dd6d68d9fbdad5a71b9/server/constants/daysOfWeek.ts#L1)). The advantage of this is that you don't need to duplicate the list anywhere. You can also render the [template easily](https://github.com/ministryofjustice/hmpps-visits-internal-admin-ui/blob/3b1606699482c8897a6e9dd6d68d9fbdad5a71b9/server/views/pages/prisons/sessionTemplates/addSessionTemplate.njk#L82-L99) from the constant too.

Not a big issue as so far this is the only place we use days of the week. If we get more uses though we'd want the approach above to avoid duplication.

server/@types/bapv.d.ts
```diff
@@ -143,6 +143,11 @@ export interface Prison extends Omit<PrisonDto, 'code'> {
   prisonId: string
 }
 
+export type PrisonRemandConfig = {
```
tpmcgowan, line 148: It's OK here but could consider exporting this from the visit allowance service as it's quite domain specific. i.e. avoiding bloat in this file (mixed results so far!)

tpmcgowan (commented):
Looking good. A few suggestions and a question about some of the content.

server/views/pages/visitAllowances/update.njk
```diff
@@ -0,0 +1,105 @@
...
+
+        {{ govukRadios({
+          name: "weekStartDay",
+          fieldset: {
+            legend: {
+              text: "On which day does the allowance renew?",
+              classes: "govuk-fieldset__legend--m"
+            }
+          },
+          items: [
```
tpmcgowan, line 55: This is fine but if you wanted to leverage your days-of-the-week constant you could do something like we have in the admin service: https://github.com/ministryofjustice/hmpps-visits-internal-admin-ui/blob/491ffa8f7a83d486ee6f422390b9d45d9a93a654/server/views/pages/prisons/sessionTemplates/addSessionTemplate.njk#L82-L99

tpmcgowan (approved):
LGTM! (optional suggestion to simplify rendering days of week but up to you)

PR #1369, opened 2026-07-03
VB-6771 Remove print visit passes feature flag

tpmcgowan:


server/utils/viewUtils.ts
```diff
@@ -5,6 +5,6 @@ const environment = new nunjucks.Environment()
 export default class ViewUtils {
   static escape(val: string): string {
     const escape = environment.getFilter('escape')
-    return escape(val).val
+    return String(escape(val))
```
tpmcgowan, line 8: Unrelated - just tweaking how this value is converted to `string`.

dhirajshettymoj (approved):
LGTM

PR #1377, opened 2026-07-09
VB-4964, VB-4966 Session blocks: show both blocked dates and sessions

tpmcgowan:
* New 'blocked dates' page to show both blocked dates and sessions
* Add API endpoint and data/service layer methods to support this
* Update integration tests

While the session blocking is still only part-implemented, the previous 'blocked dates' page is still available; controlled by the feature flag. (Various `TODO`s also added to the code to highlight what will need tidying at the end of this work.)


server/routes/blockDatesOrSessions/blockDatesOrSessionsController.test.ts
```diff
@@ -17,83 +17,211 @@ const blockDatesOrSessionsService = createMockBlockDatesOrSessionsService()
...
+
+      app = appWithAllRoutes({ services: { blockDatesOrSessionsService, visitSessionsService } })
+
       flashData = {}
       flashProvider.mockImplementation((key: keyof FlashData) => flashData[key])
     })
 
-    it('should display block visit dates page with a blocked dates listed', () => {
+    // TODO remove this test block once session blocks are fully implemented
+    describe(`GET ${url} (date blocks ONLY - sessionDateBlocks feature disabled)`, () => {
```
tpmcgowan, line 36: This is the set of tests for the old page. Basically duplicated it for the new page (currently only shown with feature flag) so these can all go once it's fully implemented.

server/routes/timetable/timetableItemBuilder.ts
```diff
@@ -74,14 +74,14 @@ const mergeGroupNames = (groupNames: string[]): string => {
 }
 
 // Function to build description of groups included/excluded from this particular session
-const buildAttendeesText = ({
+export const buildAttendeesText = ({
```
tpmcgowan, line 77: The 'Who' attendees column is supposed to have the same text as the timetable page (which is built up from session template details). So, just exporting relevant method here so it can be reused. It's already tested as part of the timetable code.

server/views/pages/blockDatesOrSessions/blockDates.njk
```diff
@@ -0,0 +1,117 @@
+{# TODO remove this template once session blocks are fully implemented #}
```
tpmcgowan, line 1: The 'old' page. Unchanged; just renamed to make space for the new one. And it will go once feature fully rolled out.

dhirajshettymoj (approved):
LGTM

PR #1379, opened 2026-07-09
VB-4966 Add choose date or session block page

tpmcgowan:
New page after selecting a date that prompts user whether they want to block a full day or a single session (only on dev where feature is enabled)

### DIFF
--- a/server/routes/blockDatesOrSessions/blockSessions/blockSessionController.ts
+++ b/server/routes/blockDatesOrSessions/blockSessions/blockSessionController.ts
@@ -1,0 +1,17 @@
+import { RequestHandler } from 'express'
+import { AuditService, BlockDatesOrSessionsService, VisitService } from '../../../services'
+
+export default class BlockSessionController {
+  public constructor(
+    private readonly auditService: AuditService,
+    private readonly blockDatesOrSessionsService: BlockDatesOrSessionsService,
+    private readonly visitService: VisitService,
+  ) {}
+
+  public view(): RequestHandler {
+    return async (req, res) => {
+      // TODO implement block session view functionality - temporary redirect to avoid 404s
+      res.redirect('/block-visit-dates/block-date-or-session')
+    }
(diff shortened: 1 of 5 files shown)


server/routes/blockDatesOrSessions/blockSessions/blockSessionController.ts
```diff
@@ -0,0 +1,17 @@
...
+  public constructor(
+    private readonly auditService: AuditService,
+    private readonly blockDatesOrSessionsService: BlockDatesOrSessionsService,
+    private readonly visitService: VisitService,
+  ) {}
+
+  public view(): RequestHandler {
+    return async (req, res) => {
+      // TODO implement block session view functionality - temporary redirect to avoid 404s
+      res.redirect('/block-visit-dates/block-date-or-session')
```
tpmcgowan, line 14: Session blocking not implemented yet (coming next!) so skeleton controller here with a redirect to avoid 404s - and it only affects dev environment.

server/routes/blockDatesOrSessions/chooseDateOrSessionBlockController.ts
```diff
@@ -1,18 +1,50 @@
...
+      }
+
+      const errors = validationResult(req)
+      if (!errors.isEmpty()) {
+        req.flash('errors', errors.array())
+        return res.redirect('/block-visit-dates/block-date-or-session')
+      }
+
+      const { blockType } = matchedData<{ blockType: 'date' | 'session' }>(req)
+      blockDateOrSession.backLinkHref = '/block-visit-dates/block-date-or-session'
```
tpmcgowan, line 39: Thinking of adding better routes/paths handling into the staff service (like we do in public service) so they're defined once and looked up instead of having so many hard-coded strings everywhere. Will file a tech debt ticket....

dhirajshettymoj (approved):
LGTM

PR #1380, opened 2026-07-10
VB-6353: Link contact page in visitor request table

sorrelingabout:
- Pull in updated typescript types from orchestration service.
- Add a new column to the visitor linking flow.

image


server/views/pages/bookerManagement/booker/visitorList.njk
```diff
@@ -40,6 +47,13 @@
       text: contact.lastApprovedForVisitDate | formatDate if contact.lastApprovedForVisitDate else "None",
       classes: "bapv-vertical-align-middle",
       attributes: { "data-test": "visitor-" + loop.index + "-last-visit" }
+    },
+    {
+      html: '<a href="'+ contactDetailsUrl +'" target="_blank" rel="noopener noreferrer">
+        View Contact <span class="govuk-visually-hidden">details for ' + contactName + ' (opens in a new tab)</span>
```
tpmcgowan:
```suggestion
        View contact <span class="govuk-visually-hidden">details for ' + contactName + ' (opens in a new tab)</span>
```

Should be lower case `'c'`. Same for this link on the other pages too!

server/views/pages/bookerManagement/booker/visitorList.test.ts
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    const $ = cheerio.load(compiledTemplate.render(viewContext))
+    const $linkElement = $('[data-test="visitor-1-contact-link"]')
+
+    expect($linkElement.text().trim()).toBe('View Contact details for Jeanette Smith (opens in a new tab)')
+
+    expect($linkElement.find('a').attr('href')).toBe(
+      'dpsContacts/prisoner/A1234BC/contacts/manage/4321/relationship/12345678',
+    )
```
tpmcgowan: We generally test Nujucks templates in tests that live alongside the relevant controller. These tests are more like integration tests really; they fire up the whole app (via `appSetup.ts`) and make HTTP requests to it and parse the returned web page HTML.

So for this booker management visitor list page the relevant test would be `visitorListController.test.ts` and you'd just need to add the assertions for your new column [here](https://github.com/ministryofjustice/book-a-prison-visit-staff-ui/blob/eea8ab20410a6e54f0269eba2e60e30b7aa6d78b/server/routes/bookerManagement/booker/visitorListController.test.ts#L73-L77).

So you don't really need this additional template-level test file. We have got a few direct template-level tests in the repo but they're more the exception - and generally for where we have (possibly unwisely!) got a lot of logic in templates.

tpmcgowan (commented):
Looks good so far. Small typo. Main comment is around testing. Happy to explain further if helpful

server/routes/testutils/testScreenReaderAndVisibleOnlyText.ts
```diff
@@ -0,0 +1,14 @@
+import { Cheerio } from 'cheerio'
+import { AnyNode } from 'domhandler'
+
+export default function testScreenReaderAndVisibleOnlyText(
+  $el: Cheerio<AnyNode>,
```
tpmcgowan:
```suggestion
import type { CheerioAPI } from 'cheerio'

export default function testScreenReaderAndVisibleOnlyText(
  $el: ReturnType<CheerioAPI>,
```

☝️ by using the above instead we can derive the type you need from Cheerio itself so there's no need to add a new dependency to the project.

package.json
```diff
@@ -90,6 +90,7 @@
     "axe-core": "^4.12.1",
     "cheerio": "^1.2.0",
     "concurrently": "^10.0.3",
+    "domhandler": "6.0.1",
```
tpmcgowan: With suggested change above to the typing within the test helper you can avoid a new dependency.

sorrelingabout: Solid, reverted these two files

server/routes/testutils/testScreenReaderAndVisibleOnlyText.ts
sorrelingabout: This makes sense! Thanks, I was struggling to work out what type would be better here. Fixed.

server/views/pages/bookerManagement/booker/visitorList.test.ts
sorrelingabout: Tests moved as suggested.

tpmcgowan (approved):
LGTM

PR #1388, opened 2026-07-15
Improve user journey for linking prisoners with no contacts

sorrelingabout:
Standard flow. No changes.
Screenshot 2026-07-15 at 16 13 39

Standard flow; no existing links (message removed).
Screenshot 2026-07-15 at 16 14 37

None of them have birthdays; existing links available. Guides admin to review.
Screenshot 2026-07-15 at 16 14 06

None of them have birthdays; no existing links. No further action can be taken. Guides admin to reject.
Screenshot 2026-07-15 at 16 14 19

No unlinked visitors. Existing links available. Guides admin to review.
Screenshot 2026-07-15 at 16 15 10

No unlinked visitors. No existing links available. Guides admin to reject.
Screenshot 2026-07-15 at 16 15 38

### DIFF
--- a/server/routes/bookerManagement/visitorRequests/visitorRequestDetailsController.test.ts
+++ b/server/routes/bookerManagement/visitorRequests/visitorRequestDetailsController.test.ts
@@ -182,7 +182,8 @@ describe('Booker management - visitor requests - link a visitor', () => {
           // Visitor list
           expect($('[data-test=no-dob-warning]').text()).toContain('must have a date of birth')
           expect($('[data-test=no-visitors-warning]').length).toBe(0)
-          expect($('[data-test=visitor-1-select] input').length).toBe(0)
+          expect($('[data-test=visitor-1-select] input').length).toBe(1)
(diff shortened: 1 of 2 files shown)


tpmcgowan (changes requested):
It looks like this is probably doing the right thing but without testing it  I'm struggling a little to reason through all this logic in the template 😵‍💫! (and I appreciate a lot of it was there already!)

However, I think we need some additional test cases (e.g. somewhere in `server/routes/bookerManagement/visitorRequests/`) to cover the two specific new scenarios on the ticket and assert that the relevant bits are / are not on the page.

Also I don't think the 'disabled' radio button for contacts without dates of birth has been done? Might need to toughen up the [relevant validation](https://github.com/ministryofjustice/book-a-prison-visit-staff-ui/blob/4d379943554bca3bbfdfc3937eb2b4d34a909fe6/server/routes/bookerManagement/visitorRequests/visitorRequestDetailsController.ts#L150-L153) to check for valid visitor ID **and** DoB.

tpmcgowan (approved):
LGTM

PR #1391, opened 2026-07-16
VB-6840 Visit session blocks: include visit room name

tpmcgowan:
Some visit sessions have the same times but are in different rooms. To disambiguate these this change adds the visit room name alongside the time slot in the session block / unblock journeys.

**Blocked dates or sessions listing page:**

Screenshot 2026-07-16 at 12 45 45

**Choose which session to block page:**

Screenshot 2026-07-16 at 12 46 24

**Success banner:**
Screenshot 2026-07-16 at 12 47 33


adaviesMOJ (approved):
LGTM :)

PR #1393, opened 2026-07-16
Tidy 'start-to-end time' formatting

tpmcgowan:
* Standardise various ways there were in the code for formatting start-to-end times, e.g. `10am to 11:30am`
* Change instances of time spans using `'-'` rather than `'to'` as a separator (in order to follow GDS guidelines)

### DIFF
--- a/server/utils/utils.ts
+++ b/server/utils/utils.ts
@@ -171,11 +171,12 @@ export const isMobilePhoneNumber = (phoneNumber: string): boolean => {
 }
 
 // E.g. 10:00, 13:30 => "10am to 1:30pm"
+// Handles either 5-digit 24H times (HH:mm format) or ISO date strings
 export const formatStartToEndTime = (startTime: string, endTime: string): string => {
   try {
     const now = new Date()
-    const startTimeAsDate = parse(startTime, 'HH:mm', now)
-    const endTimeAsDate = parse(endTime, 'HH:mm', now)
+    const startTimeAsDate = startTime.length === 5 ? parse(startTime, 'HH:mm', now) : parseISO(startTime)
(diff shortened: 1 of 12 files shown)


server/utils/utils.ts
```diff
@@ -171,11 +171,12 @@ export const isMobilePhoneNumber = (phoneNumber: string): boolean => {
 }
 
 // E.g. 10:00, 13:30 => "10am to 1:30pm"
+// Handles either 5-digit 24H times (HH:mm format) or ISO date strings
 export const formatStartToEndTime = (startTime: string, endTime: string): string => {
```
tpmcgowan, line 175: This is the main function now for doing the start-to-end time formatting. I've extended it so it handles time-only or ISO date inputs. Can be called as a Nunjucks filter too.

server/utils/utils.ts
```diff
@@ -171,11 +171,12 @@ export const isMobilePhoneNumber = (phoneNumber: string): boolean => {
...
   try {
     const now = new Date()
-    const startTimeAsDate = parse(startTime, 'HH:mm', now)
-    const endTimeAsDate = parse(endTime, 'HH:mm', now)
+    const startTimeAsDate = startTime.length === 5 ? parse(startTime, 'HH:mm', now) : parseISO(startTime)
```
psoleckimoj, line 178: Only suggestion is you could be more specific on this and use a regexp?

psoleckimoj (approved):

server/utils/utils.ts
tpmcgowan: Yeah, might be clearer what's happening if I did that. Auto merged though so will add to the next one!

PR #1397, opened 2026-07-20
VB-6834 Remove deprecated future exclude dates endpoint

tpmcgowan:
Replace with V2 endpoint that returns future date and (optionally) session blocks.


adaviesMOJ (approved):
LGTM

PR #1401, opened 2026-07-21
VB-6850 Fix double-escaping of booker visitor names

tpmcgowan:
Fix double-escaping of contact names accidentally introduced in #1380.


simonnebMOJ (approved):
LGTM

PR #1411, opened 2026-08-07
VB-6594, persist language preference

hutcheonb-moj:
Language preference was being lost (if visit had originally had CY set by the booker using the public service)
this PR persists this preference, and offers no way to change, since currently staff service doesn't support any language preference changes

### DIFF
--- a/server/routes/visitJourney/mainContact.test.ts
+++ b/server/routes/visitJourney/mainContact.test.ts
@@ -99,6 +99,12 @@ testJourneys.forEach(journey => {
         applicationReference: 'aaa-bbb-ccc',
         // visit reference only known on update journey
         visitReference: journey.isUpdate ? 'ab-cd-ef-gh' : undefined,
+        mainContact: journey.isUpdate
+          ? {
+              contactId: 123,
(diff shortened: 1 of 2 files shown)


server/routes/visitJourney/mainContact.test.ts
```diff
@@ -114,23 +120,25 @@ testJourneys.forEach(journey => {
...
-            expect($('input[name="contact"]').length).toBe(2)
-            expect($('input[name="contact"]:checked').length).toBe(0)
-            expect($('input[name="contact"]').eq(0).prop('value')).toBe('123')
-            expect($('input[name="contact"]').eq(1).prop('value')).toBe('someoneElse')
-            expect($('#someoneElseName').prop('value')).toBeFalsy()
-            expect($('#phoneNumberInput').prop('value')).toBeFalsy()
-            expect($('#email').prop('value')).toBeFalsy()
-          })
-      })
+      if (journey.isUpdate === false) {
```
adaviesMOJ, line 123: Skipping because the update journey should never be empty

adaviesMOJ (approved):

PR #1420, opened 2026-08-20
VB-6670: Add language preference checkbox to book a visit flow

sorrelingabout:
Screenshot 2026-08-20 at 11 39 19
Screenshot 2026-08-20 at 11 39 28
Screenshot 2026-08-20 at 11 39 35
Screenshot 2026-08-20 at 11 39 41
Screenshot 2026-08-20 at 11 39 56
Screenshot 2026-08-20 at 11 40 02
Screenshot 2026-08-20 at 11 40 05
Screenshot 2026-08-20 at 11 40 12

### DIFF
--- a/server/routes/visitJourney/mainContact.ts
+++ b/server/routes/visitJourney/mainContact.ts
@@ -27,6 +27,7 @@ export default class MainContact {
         ? undefined
         : visitSessionData.mainContact.contactName
       formValues.email = visitSessionData.mainContact.email ?? ''
+      formValues.languagePreference = visitSessionData.mainContact.languagePreference ?? 'en'
     }
     res.render('pages/bookAVisit/mainContact', {
       errors: req.flash('errors'),
--- a/server/routes/visitJourney/mainContact.ts
+++ b/server/routes/visitJourney/mainContact.ts
@@ -62,7 +62,7 @@ export default class MainContact {
       phoneNumber: req.body.phoneNumber === 'hasPhoneNumber' ? req.body.phoneNumberInput : undefined,
       email: req.body.email,
       contactName: selectedContact?.name ?? req.body.someoneElseName,
-      languagePreference: visitSessionData.mainContact?.languagePreference ?? 'en',
+      languagePreference: req.body.languagePreference === 'cy' ? 'cy' : 'en',
     }
 
     // update visit application to have the latest data
--- a/server/views/pages/bookAVisit/checkYourBooking.njk
+++ b/server/views/pages/bookAVisit/checkYourBooking.njk
@@ -34,6 +34,8 @@
         {%- if mainContact.relationshipDescription %} ({{ mainContact.relationshipDescription | lower}} of the prisoner){% endif %}
       {%- endset %}
 
+      {% set contactLanguageNote %} Updates in {{ "Welsh and " if mainContact.languagePreference == "cy" -}} English {%- endset %}
+
       {% set displayRows = [
         {
           key: {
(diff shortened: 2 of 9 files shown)


server/views/pages/bookAVisit/checkYourBooking.njk
```diff
@@ -34,6 +34,8 @@
         {%- if mainContact.relationshipDescription %} ({{ mainContact.relationshipDescription | lower}} of the prisoner){% endif %}
       {%- endset %}
 
+      {% set contactLanguageNote %} Updates in {{ "Welsh and " if mainContact.languagePreference == "cy" -}} English {%- endset %}
```
tpmcgowan, line 37: Stray whitespace before 'Updates in' which you've then had to include in the test assertions. (Same in confirmation.njk.). Exceedingly minor though and won't be visible

server/routes/visitJourney/mainContact.ts
tpmcgowan: Don't you need something here like in the `POST` route like:
`languagePreference: req.body.languagePreference === 'cy' ? 'cy' : 'en',`.

Route level tests in the PST block for this would then assert that the session data ended up having en / cy as appropriate depending on form data submitted.

Otherwise you're not actually using the value supplied in the checkbox? The existing line was just there to preserve/default a value.

tpmcgowan (commented):
Looks good so far. Minor comment about some whitespace and another about actually handling the submitted form value.

Would be good too to add new assertions to relevant existing route test cases where a new label has been added.

(n.b. once it's done we'll probably want to hold off merging until we've actually turned Welsh on in prod for the public service. Will happen very soon.)

server/views/pages/bookAVisit/checkYourBooking.njk
sorrelingabout: It is, but no combination of `{%-` seems to remove it.

I think this is more legible in the Nunjucks, while still allowing use of "Exact" rather than "Contains" in the tests, like other fields in this test (which is how other fields get around it, seemingly).

tpmcgowan: I mean just having `{% set contactLanguageNote %}Updates` instead of `{% set contactLanguageNote %} Updates`

server/routes/visitJourney/mainContact.ts
sorrelingabout: I believe this should be okay - We are pulling the data through correctly from the main contact to populate the form with default information.

But good spot, this has raised a slightly different bug that results in the checkbox _not_ being checked if you go back and edit the application, and it was checked to begin with.

Fixed!

server/views/pages/bookAVisit/checkYourBooking.njk
sorrelingabout: I get what you mean, but tbh I dislike it from a readability standpoint. Either we'll end up with an errant space in the tests, or "choked" text in the HTML. I don't see a way of having both be "correct" 😅 

I know it's subtle, but personally I'd rather the tests be _slightly_ out-of-expected-style than the "real" Nunjucks code

server/routes/visitJourney/mainContact.ts
tpmcgowan: Just tried this locally and it doesn't persist the en/cy choice. E.g. start a new booking. It default to unchecked (i.e. 'en') but if I check the box and continue the journey - or just go back a page - the Welsh comms preference is lost.

You need to capture the value when the form is submitted and get this into the session data - which is then ultimately sent to the back-end to set the language preference for that visit.

server/views/pages/visit/details/visitDetails.njk
```diff
@@ -55,6 +55,11 @@
             {%- endif -%}
           </dd>
 
+          <dt>Language for updates</dt>
+          <dd data-test="visit-language-preference">
```
tpmcgowan, line 59: Would be good to add a test assertion for this new list item to the existing test for this route - e.g. here https://github.com/ministryofjustice/book-a-prison-visit-staff-ui/blob/1271f8d51ab6858ce33a304c18884184413b0775/server/routes/visit/details/visitDetailsController.test.ts#L94-L103

server/views/pages/bookAVisit/checkYourBooking.njk
sorrelingabout: For some reason I forgot about `.trim()` existing...

tpmcgowan (approved):
LGTM

